PR #97, opened 2015-02-18
Implement system name override

baloo:
For automation purpose, I'd like to start lldpd with a hostname
override. `lldpcli configure system hostname` is not convenient enough
for puppet.

Signed-off-by: Arthur Gautier <baloo@gandi.net>


baloo:
No need
```
configure system hostname $hostname
```
in `/etc/lldpd.d/hostname.conf` would do the trick

PR #99, opened 2015-03-13
lldpd: implement basic port label matching for interface names

commodo:
This is a RFC for an implementation of basic port labelling
without the use of any LLDP extensions (like LLDP-MED, DOT1 & DOT3).

The way this works is:
- added a new 'portidsubtype' to the command:
   'configure portidsubtype ifname ifname | macaddress | port'
- by default 'port' is the new 'portidsubtype'
  - if there is a file present called '/etc/lldpd.labels' then
    a label is searched in that file for a given ifname
  - if there is no file, or no port label was found, the 'macaddress'
    is used as fallback
- in that file, ifnames must be matched to 'label'
  example:
    eth0=BLUE_NET
    eth1=RED_NET
    eth2=MAGIC_PORT
    ...
  Note that the parsing of this file is not very fault tolerant to spaces
  In essence, a process would set this to the file, and lldpd could be
  restarted to broadcast this info, having Port IDs that match to physical
  ports/cables/setups.
- internally, the LLDP_PORTID_SUBTYPE_PORT value is used, since it looked unused
  for desired for future use; alternative would have been to add
  a LLDP_PORTID_SUBTYPE_LABEL, or rename to this from LLDP_PORTID_SUBTYPE_PORT

Note: if chroot, be sure to put the file in the chroot location.
      example: if chroot is '/var/run/lldp', then the labels should be
               in '/var/run/lldp/etc/lldpd.labels'

I'm not looking for this to be final/official implementation that allows
us to override Port IDs to anything other than ifname and/or macaddress,
but we do need something like this.

### DIFF
--- a/src/client/conf-lldp.c
+++ b/src/client/conf-lldp.c
@@ -167,6 +167,14 @@ register_commands_configure_lldp(struct cmd_node *configure)
 				NEWLINE, NULL,
 				NULL, cmd_portid_type,
 				b_map->string);
+		} else if (!strcmp(b_map->string, "port")) {
+			commands_new(
+				commands_new(configure_lldp_portid_type,
+					     b_map->string, "Port Label/ID",
+					     NULL, NULL, NULL),
+				NEWLINE, NULL,
+				NULL, cmd_portid_type,
+				b_map->string);
 		} else if (!strcmp(b_map->string, "macaddress")) {
 			commands_new(
 				commands_new(configure_lldp_portid_type,
--- a/src/daemon/interfaces.c
+++ b/src/daemon/interfaces.c
@@ -457,6 +457,37 @@ interfaces_helper_mgmt(struct lldpd *cfg,
 	}
 }
 
+static int
+interfaces_helper_find_port_label(struct lldpd_port *port, struct lldpd_hardware *hardware)
+{
+	FILE *fp;
+	char buf[256], *p;
+	int ifname_len = strlen(hardware->h_ifname);
+
+	fp = fopen( SYSCONFDIR "/lldpd.labels", "r" );
+	if (!fp)
+		return 0;
+
+	while (fgets (buf, sizeof(buf), fp)) {
+		if ((p = strstr(buf, hardware->h_ifname)) &&
+		    (p = strstr(p + ifname_len,"="))) {
+			p++;
+			if (port->p_id != NULL) {
+				free(port->p_id);
+			}
+			if (p[strlen(p) - 1] == '\n')
+				p[strlen(p) - 1] = '\0';
+			port->p_id = strdup(p);
+			port->p_id_len = strlen(p);
+			fclose(fp);
+			return 1;
+		}
+	}
+	fclose(fp);
+
+	return 0;
+}
+
 /* Fill up port name and description */
 void
 interfaces_helper_port_name_desc(struct lldpd *cfg,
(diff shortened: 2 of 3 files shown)


vincentbernat:
I would suggest using the regular configuration file to assign labels to interface. So, the configuration command would just be:

```
configure portidsubtype eth0 port BLUE_NET
configure portidsubtype eth0 port RED_NET
configure portidsubtype eth0 port MAGIC_PORT
```

You can use a separate file in `/etc/lldpd.d` to put that information if you need to generate it from another process.

Also, the `PORT` subtype is not really for that since it should be linked with a value from the entity MIB; however, this would match perfectly the `LOCAL` subtype.

commodo:
That idea sounds pretty good.
I'll re-spin. I still have to get to understand a bit better the commands construction.

vincentbernat:
I can do that for you if you want, you just have to insert a `commands_new` to get something like that:

```c
		} else if (!strcmp(b_map->string, "port")) {
			commands_new(
                           commands_new(
				commands_new(configure_lldp_portid_type,
					     b_map->string, "Port Label/ID",
					     NULL, NULL, NULL),
                              NULL, "Port Label/ID",
                              NULL, cmd_store_env_value, "port-id"),
				NEWLINE, NULL,
				NULL, cmd_portid_type,
				b_map->string);
```

You'll be able to retrieve the value with `cmdenv_get(env, "port-id")`.

Then, you need to implement the `set_str` operation for `lldpctl_k_port_id` to be able to modify its value (so, you need a `set_str` method since it doesn't exist for `atom_port` in `atom-private.c`. And accept this change in `client.c` in the daemon.

Looking at older commits doing something similar, maybe 8fbd319539577cd2ad3ba95329dc7881849e4827 would help (it is more complex that what you want to do, I think). Maybe decaec0d8a675c0e159248e448e82807f099b2cb is a better example (but it doesn't work with port-related stuff).

commodo:
It's ok.
I'll do it myself :)

Thanks for the hints.

commodo:
Hello,

I've re-spinned the work in a separate branch: https://github.com/commodo/lldpd/commits/port_label_wip
I'm still struggling a bit with the commands.
I'd also have some questions.

Could you take a look (at the wip commit) and provide me with some hints ?
At the moment I have this command being accepted, but it does not change anything :
```
configure lldp portidsubtype local port eth0 id MGMT_PORT
configure lldp portidsubtype local port eth1 id LAN_PORT1
```
Could you provide me with some hints on what I should be looking at/doing ?
I feel like I'm close with this, and I'm only missing a few things.

I was thinking at some point to try to use the `cmd_restrict_ports()` function.
Would it make sense to use that instead of the 'port' argument ?
A related question regarding this: are `port_ids` supposed to be unique, or can multiple ports have the same ID (like in a trunking setup).
If they're supposed to be unique, then maybe, a check would be needed here to enforce uniqueness.

Maybe if I get this working, I'd probably try to allow port description override, if it's not too much work.
We don't really need it yet, so I might not add it, but I'll see.

Thanks
Alex

vincentbernat:
I'll have a look during the week.

Concerning your questions, yes, it would make sense to use `cmd_restrict_ports()`. I wouldn't try to be too smart by not allowing some configuration. Some people may need it, so if there are duplicates, let them be.

About all port-related configuration, there is currently not much in lldpd because there is a major drawback: if the port doesn't exist when the configuration is read (at start), the configuration is just ignored. Also, when applying on all ports, this only applies to existing ports. New ports won't get the configuration. I plan to fix that by implementing "ghost" ports. This is just to explain why, currently, there is not much port-restricted configuration stuff.

commodo:
Closing this, because this discussion will be moved here: https://github.com/vincentbernat/lldpd/pull/101
Keeping this here for reference.

PR #101, opened 2015-03-23
lldpd: implement basic support to override port IDs via lldpcli 

commodo:
Will reference this discussion : https://github.com/vincentbernat/lldpd/pull/99 since it's an implementation of what was discussed there.

### DIFF
--- a/src/daemon/client.c
+++ b/src/daemon/client.c
@@ -300,6 +300,11 @@ client_handle_set_port(struct lldpd *cfg, enum hmsg_type *type,
 			    port->p_id_len = strlen(set->local_id);
 			    port->p_id_subtype = LLDP_PORTID_SUBTYPE_LOCAL;
 		    }
+		    if (set->local_descr) {
+			    log_debug("rpc", "requested change to Port Description");
+			    free(port->p_descr);
+			    port->p_descr = strdup(set->local_descr);
+		    }
 #ifdef ENABLE_LLDPMED
 		    if (set->med_policy && set->med_policy->type > 0) {
 			    log_debug("rpc", "requested change to MED policy");
@@ -363,6 +368,7 @@ set_port_finished:
 	if (!ret) *type = NONE;
 	free(set->ifname);
 	free(set->local_id);
+	free(set->local_descr);
 #ifdef ENABLE_LLDPMED
 	free(set->med_policy);
 	if (set->med_location) free(set->med_location->data);
--- a/src/daemon/client.c
+++ b/src/daemon/client.c
@@ -293,7 +293,13 @@ client_handle_set_port(struct lldpd *cfg, enum hmsg_type *type,
 	TAILQ_FOREACH(hardware, &cfg->g_hardware, h_entries)
 	    if (!strcmp(hardware->h_ifname, set->ifname)) {
 		    struct lldpd_port *port = &hardware->h_lport;
-		    (void)port;
+		    if (set->local_id) {
+			    log_debug("rpc", "requested change to Port ID");
+			    free(port->p_id);
+			    port->p_id = strdup(set->local_id);
+			    port->p_id_len = strlen(set->local_id);
+			    port->p_id_subtype = LLDP_PORTID_SUBTYPE_LOCAL;
+		    }
 #ifdef ENABLE_LLDPMED
 		    if (set->med_policy && set->med_policy->type > 0) {
 			    log_debug("rpc", "requested change to MED policy");
@@ -356,6 +362,7 @@ client_handle_set_port(struct lldpd *cfg, enum hmsg_type *type,
 set_port_finished:
 	if (!ret) *type = NONE;
 	free(set->ifname);
+	free(set->local_id);
 #ifdef ENABLE_LLDPMED
 	free(set->med_policy);
 	if (set->med_location) free(set->med_location->data);
--- a/src/lib/atom-private.c
+++ b/src/lib/atom-private.c
@@ -848,6 +848,9 @@ _lldpctl_atom_set_atom_port(lldpctl_atom_t *atom, lldpctl_key_t key, lldpctl_ato
 	case lldpctl_k_port_id:
 		set.local_id = p->port->p_id;
 		break;
+	case lldpctl_k_port_descr:
+		set.local_descr = p->port->p_descr;
+		break;
 #ifdef ENABLE_DOT3
 	case lldpctl_k_port_dot3_power:
 		if (value->type != atom_dot3_power) {
@@ -1045,6 +1048,10 @@ _lldpctl_atom_set_str_port(lldpctl_atom_t *atom, lldpctl_key_t key,
 		free(port->p_id);
 		port->p_id = strdup(value);
 		port->p_id_len = strlen(value);
+		break;
+	case lldpctl_k_port_descr:
+		free(port->p_descr);
+		port->p_descr = strdup(value);
 		break;
 	default:
 		SET_ERROR(atom->conn, LLDPCTL_ERR_NOT_EXIST);
--- a/src/lib/atom-private.c
+++ b/src/lib/atom-private.c
@@ -273,6 +273,7 @@ static lldpctl_map_t bond_slave_src_mac_map[] = {
 static lldpctl_map_t lldp_portid_map[] = {
 	{ LLDP_PORTID_SUBTYPE_IFNAME,   "ifname"},
 	{ LLDP_PORTID_SUBTYPE_LLADDR,   "macaddress"},
+	{ LLDP_PORTID_SUBTYPE_LOCAL,    "local"},
 	{ LLDP_PORTID_SUBTYPE_UNKNOWN,  NULL},
 };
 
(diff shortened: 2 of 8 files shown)


commodo:
Please do not merge yet, since there is still an issue with the code.
Seems that every once in a while this crashes and I don't understand why.
I'm also looking for hints at what could be the cause.

This Pull Request is also meant to be a discussion point at first, to see if what I've implemented here is acceptable.
Regarding the local portsubtype, it seems that I cannot set that in the configuration object, since I don't have anywhere else to pull it from (as with ifname and mac portsubtypes).

commodo:
Below is a stack-trace.
Seems that it occurs when lldpd sends frames.
Any idea why this occurs only after a while ? 
It seems that it happens after a while; I am suspecting maybe when `interfaces_update()` runs ?

```
gdb) bt
#0  __GI_memcpy (to=to@entry=0x102a3e1a, from=<optimized out>, len=7) at libc/string/powerpc/memcpy.c:73
#1  0x10008d80 in lldp_send (global=0x102a1098, hardware=0x102a2808) at lldp.c:114
#2  0x100031c0 in lldpd_send (hardware=hardware@entry=0x102a2808) at lldpd.c:959
#3  0x1000bf54 in levent_send_pdu (fd=<optimized out>, what=<optimized out>, arg=0x102a2808) at event.c:731
#4  0x480251f8 in event_process_active_single_queue (activeq=0x102a1078, base=0x102a1170) at event.c:1350
#5  event_process_active (base=<optimized out>) at event.c:1420
#6  event_base_loop (base=0x102a1170, flags=flags@entry=1) at event.c:1621
#7  0x1000cb58 in levent_loop (cfg=cfg@entry=0x102a1098) at event.c:503
#8  0x10004078 in lldpd_main (argc=<optimized out>, argv=<optimized out>, envp=<optimized out>) at lldpd.c:1680
#9  0x480cdb48 in __uClibc_main (main=0x100017f0 <main>, argc=5, argv=0xbf9bb8e4, app_init=0x100017cc <_init>, app_fini=0x1000dc0c <_fini>, rtld_fini=<optimized out>, 
    stack_end=<optimized out>) at libc/misc/internals/__uClibc_main.c:486
```

src/lib/atom-private.c
```diff
@@ -1025,6 +1029,33 @@ _lldpctl_atom_get_str_port(lldpctl_atom_t *atom, lldpctl_key_t key)
...
+	    (struct _lldpctl_atom_port_t *)atom;
+	struct lldpd_port     *port     = p->port;
+
+	if (!value || !strlen(value))
+		return NULL;
+
+	switch (key) {
+	case lldpctl_k_port_id:
+		if (port->p_id != NULL) {
+			free(port->p_id);
```
vincentbernat: `free` is already checking if the value is `NULL`, you don't need to check again.

src/daemon/client.c
```diff
@@ -293,7 +293,18 @@ client_handle_set_port(struct lldpd *cfg, enum hmsg_type *type,
 	TAILQ_FOREACH(hardware, &cfg->g_hardware, h_entries)
 	    if (!strcmp(hardware->h_ifname, set->ifname)) {
 		    struct lldpd_port *port = &hardware->h_lport;
-		    (void)port;
+		    if (set->p_id && strlen(set->p_id) > 0) {
+			    log_debug("rpc", "requested change to Port ID");
+			    if (port->p_id) {
+				    free(port->p_id);
```
vincentbernat: Same here, just use `free` directly.

src/daemon/client.c
```diff
@@ -293,7 +293,18 @@ client_handle_set_port(struct lldpd *cfg, enum hmsg_type *type,
...
+			    }
+			    port->p_id = strdup(set->p_id);
+			    port->p_id_len = strlen(set->p_id);
+			    port->p_id_subtype = LLDP_PORTID_SUBTYPE_LOCAL;
+		    } else {
+			    /* This is the default value */
+			    port->p_id_subtype = LLDP_PORTID_SUBTYPE_LLADDR;
```
vincentbernat: You should do this only if `p_id` is not `NULL` (maybe when it is empty just empty). Otherwise, we can't know if that's really the port ID that we want to change. Any other port-related change would reset the port subtype to link-layer address.

src/lldpd-structs.h
```diff
@@ -275,6 +275,7 @@ MARSHAL_END(lldpd_port);
 /* Used to modify some port related settings */
 struct lldpd_port_set {
 	char *ifname;
+	char *p_id;
```
vincentbernat: I would call this new variable `local_id` as we expect a NULL-terminated string and we assume this is a local ID. Otherwise, it may be confusing in the future.

src/lib/atom-private.c
vincentbernat: Funny fact, I just found other occurrences of this in my code. ;-)

src/daemon/client.c
vincentbernat: Moreover, setting the type back to LLADDR is not enough since the value is incorrect. Since currently, the code cannot be hit, just don't do anything if the provided local ID is empty.

src/lib/atom-private.c
commodo: Actually, it's the reason I added the NULL checks [since I took them as reference].
I wasn't sure why they were there; I assumed that maybe some other libc implementation of free() does not NULL check.
Anyway; will replace these with simple free() calls.

vincentbernat:
About the segfault, you are right: in `interfaces.c`, in `interfaces_helper_physical()`, we call `lldpd_port_cleanup()` to remove any previously collected information (in case this is not valid anymore). An easy fix would be to modify `lldpd_port_cleanup()`:

```c
if (all) {
  free(port->p_id);
  port->p_id = NULL;
}
```

As for your last sentence, I didn't understand it. Could you elaborate?

src/lib/atom-private.c
vincentbernat: All POSIX implementations of `free()` should do nothing if the provided pointer is `NULL`: http://pubs.opengroup.org/onlinepubs/009695399/functions/free.html

I'll try to remove the useless `if` in the current code.

commodo:
I'm assuming it's regarding this sentence that I need to elaborate
>Regarding the local portsubtype, it seems that I cannot set that in the configuration object, since I don't have anywhere else to pull it from (as with ifname and mac portsubtypes).


When setting `configure [ports ....] lldpd portidsubtype ifname | macaddress`, the portidsubtype is set to the `c_lldp_portid_type` field.
Regarding the `local` port-subtype, I am doing things a bit differently; I am avoiding that `c_lldp_portid_type` field, because if I set it, I won't have anywhere to read the port_id from in `interfaces_helper_port_name_desc()`.

* When `cfg->g_config.c_lldp_portid_type == LLDP_PORTID_SUBTYPE_IFNAME`  interface names are used
* When `cfg->g_config.c_lldp_portid_type == LLDP_PORTID_SUBTYPE_LLADDR`  MAC addresses are used.
* When `cfg->g_config.c_lldp_portid_type == LLDP_PORTID_SUBTYPE_LOCAL` I would need a file or something from where to read port_ids.
Or, an alternative is to set them per-port as I did; however I don't know yet how to revert (if needed) back to `ifname` or `macaddress` if needed; for my case specifically I don't need to, so I'm fine.

One note is : the configure command supports that `ports` parameter.
`configure [ports ....] lldpd portidsubtype ifname | macaddress`
The `ports` param has no effect for the `ifname` and `macaddress` parameter, and for the `local` argument the `ports` has effect; but I guess if that's documented in the wiki, it shouldn't matter much.

One more question: is it ok if I force-push to this branch ?
Or should I add just add commits on top ?
Doing force-push will reset some comments, since the history will be re-written.

commodo:
And thanks a lot for the help so far :)

vincentbernat:
❦ 24 mars 2015 00:18 -0700, Alexandru Ardelean <notifications@github.com> :

> * When cfg->g_config.c_lldp_portid_type == LLDP_PORTID_SUBTYPE_LOCAL I
>   would need a file or something from where to read port_ids. Or, an
>   alternative is to set them per-port as I did; however I don't know
>   yet how to revert (if needed) back to ifname or macaddress if
>   needed; for my case specifically I don't need to, so I'm fine. 

We can work on reverting later.

> One note is : the configure command supports that ports parameter.
> configure [ports ....] lldpd portidsubtype ifname | macaddress
> The ports param has no effect for the ifname and macaddress parameter,
> and for the local argument the ports has effect; but I guess if that's
> documented in the wiki, it shouldn't matter much.

Yes, that can be fixed later by introducing a pseudo default port from
which new ports are derived.

> One more question: is it ok if I force-push to this branch ?
> Or should I add just add commits on top ?
> Doing force-push will reset some comments, since the history will be
> re-written.

Yes, force push, I prefer not to have "fix" commits.
-- 
Make the coupling between modules visible.

commodo:
So, with the success on port ID override, I decided to give another try at port description.
If it's ok with you I would like to also add it.

If not, we can drop it, or do another PR.

Thanks

commodo:
@vincentbernat ping :)

vincentbernat:
ENOTIME. I'll have a look this week-end.

commodo:
No worries.
I just wanted a rough estimate (like a day, a week, a month), in order to plan my next stuff to do.

Thanks

vincentbernat:
I am adding a few commits. Check them out.

vincentbernat:
Also, it would be nice to be able to set port description without setting port ID. But I don't need it and if you don't need it, we can just wait for someone to ask for it.

commodo:
Right now the overridable Port IDs is required.
The ability of overriding port descriptions is optional, but a nice touch.
So, what's done now, is perfect.

I'll backport some of your patches into our tree.

Merci beaucoup :)

PR #106, opened 2015-04-09
Make some TLVs optional

commodo:
Specifically, the Chassis Capabilities and Management IP addresses.
For our case they are not needed.

This change allows lldpcli to disable them by calling
```
unconfigure lldp capabilities-advertisements
unconfigure lldp management-addresses-advertisements
```

Any comments/objections/preferences are welcome for changing some namings.

### DIFF
--- a/src/client/client.h
+++ b/src/client/client.h
@@ -128,7 +128,7 @@ void register_commands_watch(struct cmd_node *);
 /* conf*.c */
 void register_commands_configure(struct cmd_node *);
 void register_commands_configure_system(struct cmd_node *, struct cmd_node *);
-void register_commands_configure_lldp(struct cmd_node *);
+void register_commands_configure_lldp(struct cmd_node *, struct cmd_node *);
 void register_commands_configure_med(struct cmd_node *, struct cmd_node *);
 void register_commands_configure_dot3(struct cmd_node *);
 void register_commands_medpow(struct cmd_node *);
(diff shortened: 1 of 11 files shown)


commodo:
The unit tests seem to have failed for `check_lldp` because I am using the global field inside of `lldp_send()`.
So, then a question: should/can I use that global field or not ?

If I can use it, then I'll update the unit test with a default `struct lldpd_config` object.
If not, then I'll update the default behaviour of the option settings.

vincentbernat:
I would put that under `unconfigure lldp` instead of `unconfigure system`.

vincentbernat:
Yes, you can use the global field.

commodo:
Re-spin complete with passing tests [Travis + my local testing I'm using].

commodo:
Is there a time frame when this would get in ?
Or is there anything else I should change to this PR ?

I'm kind of ready with the custom TLVs logic and I'd like to start the discussion with that.

Thanks :)

vincentbernat:
Sorry, I am currently swamped. I need at least 15 days to get back on tracks.

commodo:
No problem.
I can adapt our tree with patches from my changesets and when you'll be updating/releasing I'll be updating our tree.

I won't be bothering you until you're ready.

Thanks :)

vincentbernat:
Looks fine!

PR #113, opened 2015-05-21
lldpd: consider interfaces which are not running also for RTM_DELLINK event

commodo:
That way hardware->h_flags get updated, and the hardware object
is not free'd.
When the interface comes back online (with a link), the iface->flags
(and consequently hardware->h_flags) get updated.

This is a bit easier than trying to do other more complex logic,
for keeping duplicate (persistent) information.

### DIFF
--- a/src/daemon/netlink.c
+++ b/src/daemon/netlink.c
@@ -131,11 +131,6 @@ netlink_parse_link(struct nlmsghdr *msg,
     ifi = NLMSG_DATA(msg);
     len = msg->nlmsg_len - NLMSG_LENGTH(sizeof(struct ifinfomsg));
 
-    if (!((ifi->ifi_flags & IFF_UP) && (ifi->ifi_flags & IFF_RUNNING))) {
-        log_debug("netlink", "skip down interface at index %d",
-          ifi->ifi_index);
-        return -1;
-    }
     if (ifi->ifi_type != ARPHRD_ETHER) {
         log_debug("netlink", "skip non Ethernet interface at index %d",
           ifi->ifi_index);
@@ -303,6 +298,7 @@ netlink_recv(int s,
                 log_debug("netlink", "received end of dump message");
                 end = 1;
                 break;
+            case RTM_DELLINK:
             case RTM_NEWLINK:
                 if (!ifs) break;
                 log_debug("netlink", "received link information");


commodo:
@vincentbernat 
This is more of a proposal.
I remember you saying something about ghost ports, but for our case, this helps now that is implemented.
Not sure what 'ghost ports' would mean, but it sounds like duplicating info ?

In any case, we'll keep this patch in our tree until a better solution comes along.

[Explanation]
This is (quick) fix for cases where an interface goes ifup/ifdown/ifup on the other side of the cable.
Then, custom TLVs and local port ID + descriptions get reset.
One solution would be to re-run `lldpcli` on the original configs, though that could bring up more issues than we'd like right now.

vincentbernat:
Ghost ports would be ports that were present but not anymore but for which we have a special configuration set. This would matches your use case if this was implemented (but this is not the case).

I am not applying the patch now because I have a pending patch for the whole interface stuff which is waiting since quite some time and I would like to apply it first.

commodo:
Sure, no problem.
If you're ok with this patch, I'll rebase + conflict resolution + re-submit

commodo:
Closing this.
We've found some issue with this, where it gets annoying.

lldpd's netlink event handling on RTM_DELLINK is no longer useful for clearing stale states/info.
Will re-spin this logic internally, and if it's upstreamable will come back with another PR.

Thanks :)
Alex

vincentbernat:
I am not very good at following up pull requests correctly and I did forget to update this one. I still have a pending patch for the interface stuff but it should be less invasive than it was. I also think I'll be able to implement "ghost" ports during the next month.

commodo:
No worries.
I am also swamped atm; have to integrate lots of stuff and use LLDP as a basis for communicating data between ports.
I am eager to wait on that ghost ports feature ; it will simplify things a lot on our side.

With the custom TLVs support we've managed to send some internal data between ports that makes our GUIs light up like Christmas trees, so that was a big hit [and may become even bigger] in our company; so, thanks a lot for the help on that :).

vincentbernat:
Hey Alexandru!

I have pushed an important change in lldpd to use libnl3, notably
because of its caching feature. So, I'll move to ghost ports next.

commodo:
I saw that yesterday, while checking the sources.
Thanks a lot :)

vincentbernat:
There is also now a "default" port. I'll try to avoid ports going down to lose their configuration.

PR #114, opened 2015-05-21
Add -Wdocumentation (Clang CFLAG) and fix warning

alagoutte:
Add also -Wheader-guard CLFAG


vincentbernat:
The warnings with third-party headers are annoying. Is there a way to restrict to a directory?

alagoutte:
Fix upstream ? ;-)

It is possible to use "pragma" for disable for some include

http://stackoverflow.com/questions/20078941/how-to-use-pragma-clang-diagnostics

vincentbernat:
That seems a bit overkill. I gladly take the first two patches but not the last one. Is that OK for you?

alagoutte:
Like you prefer

i have start to work on using pragma on my clang2 branch (no yet fix all include...)
What do you think ? 
May be add some macro ?

vincentbernat:
I'd like not to clutter source with that. It depends how much lines you add. Moreover, currently, there is no real documentation that could be generated from lldpd code.

alagoutte:
I have make the change on clang2 branch ( https://github.com/alagoutte/lldpd/tree/clang2 )

It is good for you ? (need to rebase)

vincentbernat:
❦ 22 mai 2015 02:01 -0700, Alexis La Goutte <notifications@github.com> :

> I have make the change on clang2 branch (
> https://github.com/alagoutte/lldpd/tree/clang2 )
>
> It is good for you ? (need to rebase)

OK, that's small enough.

For interfaces-linux.c and priv-linux.c, I would rather enclose all
"linux/*" headers in the pragma: we don't really care about errors in
third-party headers.
-- 
Clothes make the man.  Naked people have little or no influence on society.
		-- Mark Twain

alagoutte:
it will be good !

vincentbernat:
Thanks!

PR #115, opened 2015-05-28
Don't use container-based Travis CI

alagoutte:
When i have fork the project and try to build on Travis,

Travis try to use container based (and no old travis infra) and the code don't build 

(May be a another solution, it is to try to use new container infra...)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: c
 os: linux
 compiler: gcc
+sudo: require
 install:
   - ./.travis/install.sh
 script:


vincentbernat:
I don't quite understand. Do you have a link to a failed build?

alagoutte:
You can look https://travis-ci.org/alagoutte/lldpd/builds/64433241

vincentbernat:
Oh, OK, the message is pretty straightforward and there is little workaround for this.

PR #119, opened 2015-06-23
redhat: fix minor issues in .spec file

kcgthb:
- fix BuildRequires package name (pkgconfig instead of pkg-config)
- fix installation source for bash_completion file

### DIFF
--- a/redhat/lldpd.spec
+++ b/redhat/lldpd.spec
@@ -81,7 +81,7 @@ protocol. It also handles LLDP-MED extension.
 %package devel
 Summary:  Implementation of IEEE 802.1ab - Tools and header files for developers
 Group:    Development/Libraries/C
-BuildRequires: pkg-config
+BuildRequires: pkgconfig
 Requires: lldpd = %{version}-%{release}
 
 %description devel


djzort:
it may be prudent to create different .spec files for different rpm distros, as redhat loves to change tiny things

kcgthb:
I agree. Although in that case:
* `pkgconfig` has been called `pkgconfig` in Red Hat since at least RHEL2
* the Makefile in `src/client` installs the bash_completion.d file in `%{_datadir}` (/usr/share) instead of `%{_sysconfdir}` (/etc) since b6371c9. I guess it's been done for Debian, but it doesn't really apply for RHEL-based distros, AFAIK.

vincentbernat:
For `pkgconfig` vs `pkg-config`, I thought that either SuSE or RHEL was able to fix that automatically. Currently, OpenSuSE Build Service is able to build the package as is for RHEL >= 4 and SuSE 11 and more. Maybe the "fix" is done automatically by OBS. Let's change that and see what happens at the next build.

For bash completion, the change from `/etc/bash-completion.d` to `/usr/share/bash-completion` was done after 0.7.15. Do you say that having completion in `/usr/share/bash-completion` won't work with RHEL? In this case, the appropriate location should also be put in `./configure` call, otherwise, this will break on next version. Or do you say that when compiling 0.7.15 with the spec file from `HEAD`, the completion is not at the right place? In this case, no fix needed, you just need to use a spec file before 98bf70e3d133f.

kcgthb:
Salut Vincent !

Indeed in OpenSuSE, the package is called `pkg-config`, so I guess the build service does the translation automatically.

For bash completion, from the FAQ in https://anonscm.debian.org/cgit/bash-completion/bash-completion.git/plain/README, the recommended way seems to be installing completions in `completiondir`, which defaults to `/etc/bash_completion.d`. They give an example on how to do that with autotools. 
On RHEL6.6 at least, the generated Makefile installs the completion file in `/etc/bash_completion.d`, but the .spec file expects it in `/usr/share`

vincentbernat:
❦ 23 juin 2015 14:48 -0700, Kilian Cavalotti <notifications@github.com> :

> On RHEL6.6 at least, the generated Makefile installs the completion
> file in /etc/bash_completion.d, but the .spec file expects it in
> /usr/share

You mean the one from lldpd? This will be changed in the next
version. If you are using the .spec from git on 0.7.15, you need to use
the .spec matching this version:
 https://github.com/vincentbernat/lldpd/blob/0.7.15/redhat/lldpd.spec

What I didn't test is that bash completions work when put in
/usr/share/bash-completion on RHEL. I made the changes because it was
required for Debian but maybe this is a Debian-only stuff. In this case,
I'll revert this change and put back completion in
/etc/bash-completion.d for everyone (except Debian).
-- 
Use recursive procedures for recursively-defined data structures.

kcgthb:
Yup, I meant the Makefile from lldpd.
But I re-checked, and it indeed looks like the source tarball that was used to build the RPM was the one from the 0.7.15 release, not the tarball generated from `make dist`. So sorry about the noise.

And BTW, great job on lldpd, but cleaner interface and much easier to use that lldpad!

vincentbernat:
Thanks! ;-)

Would you like to edit your commit for the pkg-config stuff?

kcgthb:
Sure, done.

vincentbernat:
Thanks!

PR #124, opened 2015-07-30
lldpcli: rename "unconfigure lldp custom-tlvs" to "unconfigure lldp custom-tlv"

commodo:
There was a bit of confusion regarding this command.
It's documented as 'unconfigure lldp custom-tlv', when in fact it's 'unconfigure lldp custom-tlvs' (plural).

That's also my fault, since I did not update much (or take care about) documentation.
It seems simpler to rename the command as is in the doc than change the doc.

### DIFF
--- a/src/client/conf-lldp.c
+++ b/src/client/conf-lldp.c
@@ -454,7 +454,7 @@ register_commands_configure_lldp(struct cmd_node *configure,
 	register_commands_configure_lldp_custom_tlvs(configure_lldp);
 	commands_new(
 		commands_new(unconfigure_lldp,
-			"custom-tlvs",
+			"custom-tlv",
 			"Clear all (previously set) custom TLVs",
 			NULL, NULL, NULL),
 		NEWLINE, "Clear all (previously set) custom TLVs",


vincentbernat:
We could update the documentation instead, since `custom-tlv` also match `custom-tlvs`, so nothing would break.

commodo:
I don't mind either approach.
At the moment, we don't use the `unconfigure lldp custom-tlv(s)` command, since we configure custom LLDPs at startup and don't modify them (yet).

I think I will also implement [in the short future] a  `unconfigure lldp custom-tlv subtype XX`  version to be able to unconfigure custom TLVs on-the-fly via liblldpctl.

I think that now, whether it's  `unconfigure lldp custom-tlv` or  `unconfigure lldp custom-tlvs` (plural) is fine (whichever it is), since it's still early.

Maybe the `unconfigure lldp custom-tlv` version would be best for 2 reasons:
* it's documented as such
* [as I saw it] most configure + unconfigure commands have the same name 

In any case, I leave the decision up to you.

Thanks

vincentbernat:
Ok for the second reason!

PR #130, opened 2015-08-12
Make the default platform name really the kernel name

bapt:
Instead of hardcoding 'Linux" directly query the kernel name of the user
is not overwriting it.


vincentbernat:
Thanks! I have merged this as fc37317f7cd2 (slight modification to prefix the description) and I have amended another commit.

PR #142, opened 2015-10-07
Tiny typo in lldpd's manpage

matsimon:
Just stumbled upon when reading it, I'd say stick to singular since that's what is used 2 in 3 times.


vincentbernat:
Thanks!

PR #160, opened 2016-02-10
netlink: fix reception of large netlink messages

davidmorel:
If received netlink message is bigger than the buffer provided by the
application, netlink message flag will contain the MSG_TRUNC flag. In
this case it is up to the application to allocate a buffer large enough
to receive the full message and to call recv() again.

-------------

This patch is a rework of a patch we made on top of 0.7.16, netlink handling has been changed in lldpd since then, the function changed has added labels/goto, it should not break anything but I felt it was worth mentionning, to ease the review process.

We tested this against a machine containing 26 NICs without the patch, the neighbors won't receive any lldp update, with the patch they do arrive. A quick internal test plan has been passed too.

### DIFF
--- a/src/daemon/netlink.c
+++ b/src/daemon/netlink.c
@@ -360,8 +360,9 @@ netlink_recv(int s,
     struct interfaces_device_list *ifs,
     struct interfaces_address_list *ifas)
 {
-	char reply[NETLINK_BUFFER] __attribute__ ((aligned));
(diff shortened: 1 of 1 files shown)


davidmorel:
I'll rework it to fix tests.

davidmorel:
Will redo with Organization repository, closing for now.

PR #161, opened 2016-02-10
netlink: fix reception of large netlink messages

davidmorel:
If received netlink message is bigger than the buffer provided by the
application, netlink message flag will contain the MSG_TRUNC flag. In
this case it is up to the application to allocate a buffer large enough
to receive the full message and to call recv() again.

---

This fix is a rework of our patch on 0.7.16, netlink handling code has changed since then, among other things, label/goto has been added in the same function. I checked this doesn't look like breaking anything, but I felt it was worth mentioning so it may be looked at closely in the review process.

This has been tested on a board with 26 NICs where LLDP frames were not sent anymore without the patch.

Additionally, a quick test plan was followed to check normal behavior was working fine.

### DIFF
--- a/src/daemon/netlink.c
+++ b/src/daemon/netlink.c
@@ -360,8 +360,9 @@ netlink_recv(int s,
     struct interfaces_device_list *ifs,
     struct interfaces_address_list *ifas)
 {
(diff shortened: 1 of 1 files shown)


vincentbernat:
Looks fine for me. Thanks!

PR #163, opened 2016-02-21
enhance custom tlv support

kanrajag:
support optional add / replace commands during custom tlv creation. if no option is specified it will default to "add"
support optional deletion of specific custom tlv based on oui and subtype
updated man page

### DIFF
--- a/src/lib/lldpctl.h
+++ b/src/lib/lldpctl.h
@@ -789,6 +789,7 @@ typedef enum {
 	lldpctl_k_custom_tlv_oui,		/**< `(I,WO)` custom TLV Organizationally Unique Identifier. Default is 0 (3 bytes) */
 	lldpctl_k_custom_tlv_oui_subtype,	/**< `(I,WO)` custom TLV subtype. Default is 0 (1 byte) */
 	lldpctl_k_custom_tlv_oui_info_string,	/**< `(I,WO)` custom TLV Organizationally Unique Identifier Information String (up to 507 bytes) */
+	lldpctl_k_custom_tlv_op,		/**< `(I,WO)` custom TLV operation */
 
 } lldpctl_key_t;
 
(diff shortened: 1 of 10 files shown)


src/lib/lldpctl.h
```diff
@@ -786,6 +786,7 @@ typedef enum {
 	lldpctl_k_custom_tlvs = 5000,		/**< `(AL)` custom TLVs */
 	lldpctl_k_custom_tlvs_clear,		/** `(I,WO)` clear list of custom TLVs */
 	lldpctl_k_custom_tlv,			/** `(AL,WO)` custom TLV **/
+	lldpctl_k_custom_tlv_op,		/**< `(I,WO)` custom TLV operation */
```
vincentbernat: You need to put this one at the end of the "section". Otherwise, you introduce an ABI incompatibility (existing apps will break).

vincentbernat:
This looks good to me. Maybe @commodo could have a look too.

See the inline comment for the position of `lldpctl_k_custom_tlv_op`. If you could also add an example to `tests/lldpcli.conf` (I run those from time to time).

commodo:
Looks good from my side.
I haven't run the code though yet ; if you'd want me to do that, then on Monday.

I look forward to installing it on our devices.
The `remove` command would allow us to rewrite some bits.

@kanrajag  Thanks a lot !!!  :)

kanrajag:
@vincentbernat - I did the changes you asked. please review

@commodo - glad to help out! btw - the "remove" op is via "unconfigure" option.

vincentbernat:
I don't see any change. I see that you pushed an amended commit but there is no change in it. You can just add an additional commit. I'll squash them myself.

vincentbernat:
Perfect, thanks!

PR #168, opened 2016-03-13
Replace integration tests with namespaces + py.test

vincentbernat:


vincentbernat:
Partially merged without the Travis part. tmpfs doesn't seem to be present in Travis kernels. Don't bother investigating more for now, will try to run tests in Docker instead.

PR #174, opened 2016-04-21
Fix size of PEEK_DISCARD_UINT32()

jonasj76:
Found this typo.
The macro is not used anywhere but perhaps in the future, so better get it right :)


vincentbernat:
Thanks. Yes, better safe than sorry! :)

PR #175, opened 2016-04-22
Make build of unit tests optional

jonasj76:
This commit will adds a `configure` option, `{with, without}-unittests` (default: yes),
which can be used to disable building the unit tests.

In my case, rebuilding `lldpd` often, disabling the build of unit tests shorten the building process a little bit.
I also identified a problem when cross-compiling, if `check` could be found on the host system but not for the target system, the compilation failed due to that `configure` found `check` (on host system) and then tried to use it when building the target system. I might have done something wrong in my setup, but disabling the unit test solved my problem.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -200,9 +200,6 @@ m4_popdef([AC_LANG_CALL(C)])
 
 AC_CACHE_SAVE
 
-## Unit tests wich check
-PKG_CHECK_MODULES([CHECK], [check >= 0.9.4], [have_check=yes], [have_check=no])
-
 # Third-party libraries
 lldp_CHECK_LIBEVENT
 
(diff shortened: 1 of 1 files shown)


vincentbernat:
Unittests are not built nor run if you don't do an explicit `make check`. Don't you have something like that in your build system?

jonasj76:
Sorry, I was to quick in my assumption. A closer look shows that, as you say, the unit tests are not built.
But two files in the `tests` directory are built, `common.c` and `decode.c`, where `common.c` are including `check.h`, which in my case does not exists on my cross-compile target system.

vincentbernat:
❦ 22 avril 2016 09:10 -0700, Jonas Johansson <notifications@github.com> :

> Sorry, I was to quick in my assumption. A closer look shows that, as
> you say, the unit tests are not built.
> But two files in the tests directory are built, common.c and decode.c,
> where common.c are including check.h, which in my case does not exists
> on my cross-compile target system. 

OK, I see. I'll just try to find a way to not build them outside of make
check.
-- 
Identify bad input; recover if possible.

jonasj76:
Ok, thanks.

PR #177, opened 2016-04-29
Fix AppArmor "no" option in configure script

jonasj76:
Directory for AppArmor profiles is always set due to a typo in `configure.ac`.


vincentbernat:
Thanks!

PR #184, opened 2016-06-10
seccomp: add fcntl and getsockname to seccomp whitelist

chutz:
Recent versions of lldpd make calls to fcntl and getsockname, but they
are not in the seccomp whitelist. This patch adds them. Reported by Gentoo
users in these bugs:

https://bugs.gentoo.org/577576
https://bugs.gentoo.org/564566


vincentbernat:
It has been a long time since fcntl() and getsockname() are used. I suppose that previously, they were implicitely authorized. Thanks for the patch!

PR #191, opened 2016-07-28
[RFC] lib/connection: simplify lldpctl_recv() function

commodo:
The final aim is to have async IO.

We're using uloop() (from OpenWrt) which is a wrapper on top
of epoll().
uloop() is a main-loop type of mechanism, which means FDs need
to be assigned to it, and then if the epoll() has some events,
it would trigger the handlers for the registered (with epoll) FDs.

With the current way liblldpctl looks like, async IO is not too
perfect.

But, regarding lldpctl_recv() it looks like it could call the
conn->recv() pointer, since it's mandatory anyway to pass it
in order to get a async-IO-ish mechanism.

### DIFF
--- a/src/lib/connection.c
+++ b/src/lib/connection.c
@@ -27,6 +27,9 @@
 #include "../ctl.h"
 #include "../log.h"
 
+static ssize_t
+__lldpctl_recv(lldpctl_conn_t *conn, const uint8_t *data, size_t length);
+
 const char*
 lldpctl_get_default_transport(void)
 {
@@ -170,7 +173,7 @@ _lldpctl_needs(lldpctl_conn_t *conn, size_t length)
 		free(buffer);
 		return SET_ERROR(conn, LLDPCTL_ERR_EOF);
 	}
-	rc = lldpctl_recv(conn, buffer, rc);
+	rc = __lldpctl_recv(conn, buffer, rc);
 	free(buffer);
 	if (rc < 0)
 		return SET_ERROR(conn, rc);
@@ -231,8 +234,8 @@ end:
 	return (rc);
 }
 
-ssize_t
-lldpctl_recv(lldpctl_conn_t *conn, const uint8_t *data, size_t length)
+static ssize_t
+__lldpctl_recv(lldpctl_conn_t *conn, const uint8_t *data, size_t length)
 {
 
 	RESET_ERROR(conn);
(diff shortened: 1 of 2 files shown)


commodo:
To add to the git commit comment.
I wanted to ask how to do async IO with liblldpctl ; after looking into it, it seems that there is some stuff started on supporting that.

Maybe I missed it ; is there an example of how to do that ?
So, far the async-IO with uloop() works reasonably ok, but it's not perfect.

I"ve implemented some async-ish send & recv funcs.
The reason I say they're async-ish is because `_lldpctl_needs()` does not seem to work well, when it cannot read the amount of bytes it expects.
So, I've made them use `poll()` until data is received.
Writing seems to be fairly ok ; reading is a bit more difficult ; seems to interfere a bit with `_lldpctl_do_something()`

vincentbernat:
I already have a successful integration into libevent. I should publish an example, maybe as a web server. There is another example available here: https://github.com/baloo/leaf/blob/master/src/lldp.c.

The model used is the same as for OpenSSL. The library will signal with error messages if it needs to receive more bytes. If `_lldpctl_needs()` doesn't have enough bytes, it will signal it and the signal will be propagated to the top. So, if you call `lldpctl_get_interfaces()` and you receive `LLDPCTL_ERR_WOULDBLOCK`, you need to retry later (when your event loop provides more data for the connection).

In retrospective, a callback model would have been more convenient.

commodo:
@vincentbernat build failure is related to build-system ? or ?

commodo:
hmm ; ok ; will try that ;
regarding this patch, opinions ?

vincentbernat:
From what I understand, the patch may make it randomly work if only one receive is needed to get the data. It will fail if the data needs more reads. If you have a git repository somewhere with your code, I could have a look (but currently, not really the right time, I am too short on time).

For the build problem, yes, likely a problem related with Travis.

commodo:
Ah, no need to bother looking into this much.
I am happy with the example (baloo/leaf) to use as base.
Actually, what I did resembles baloo/leaf a bit.

I wanted my patch to prompt a question of example for async.
And from a general/high-level view, it also seemed to make sense to have `lldpctl_recv()` resemble `lldpctl_send()` a bit more in terms of parameters and maybe behavior.

So, I'll close this for now.

Thanks
Alex

PR #198, opened 2016-08-30
lib/atoms/port: fix segfaults with minimal information

The-42:
When no packet has yet been received, querying LLDP information leads to
segfaults. Fix those by adding a couple pointer checks.

Signed-off-by: Bert van Hall <bert.vanhall@avionic-design.de>

### DIFF
--- a/src/lib/atoms/port.c
+++ b/src/lib/atoms/port.c
@@ -156,10 +156,12 @@ _lldpctl_atom_new_port(lldpctl_atom_t *atom, va_list ap)
 	if (port->parent)
 		lldpctl_atom_inc_ref((lldpctl_atom_t*)port->parent);
 
-	/* Internal atom. We are the parent, but our reference count is not
-	 * incremented. */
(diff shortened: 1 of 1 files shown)


vincentbernat:
Thanks! Were you running into this bug because you requested information on an interface without querying the list of interfaces?

The-42:
You're welcome :)

The library is used in a bigger application which I was about to test when I ran into this issue. The interfaces are queried before any data is then gathered. The switch the device is connected to doesn't send much, in total a mere 34 Bytes about every 30 seconds. Things worked fine, unless *lldpd* was restarted right before querying. Raw data consisted only of 20 Bytes, and the port TLV was completely missing, making the lib behave a bit... segfaulty. When the next packet arrived things worked fine again. I was unsure how and where those remaining 20 Bytes were cached, so I took the most obvious road to pointer checks and it worked out well enough.

There may be more to it though, I cannot claim to have fully traced the data through the LLDP library.

PR #203, opened 2016-10-06
client: add an option to use pre-0.9.2 json-c format

vincentbernat:
The format has been changed in c8b8b858bbba to match the JSON format
used when compiled with Jansson. Some users may want to revert this
change. Add `--enable-json0` option for that.


kanrajag:
i verified the json-c part and it looks OK. i am not sure why you modified the jansson writer as well. i have not tested that part. please go ahead with commit. thanks!

PR #212, opened 2016-11-30
linux: add support for ethtool GLINKSETTINGS

vincentbernat:
With a recent enough kernel, one can use GLINKSETTINGS instead of
GSET. This is needed to overcome the limitation on the number of
supported/advertising modes supported by GSET (which is a fixed-length
bitset). We use the same kind of compatibility structure than for
ethtool.

Also, update the list of supported MAU types to support higher speed
interfaces. Unfortunately, while Linux reports the exact
advertised/supported modes, the operational mode of the interface is
only reported with its speed and its kind (copper, fiber, ...). When
this is fiber, we don't know if this is MMF or SMF or the supported
distance. Therefore, we use our best guess (this was already the case,
nothing new here).


src/daemon/priv-linux.c
```diff
@@ -54,28 +55,88 @@ priv_open(char *file)
...
+		uint32_t link_mode_data[3 * ETHTOOL_LINK_MODE_MASK_MAX_KERNEL_NU32];
+	} ecmd;
+	static int8_t nwords = 0;
+
+	if (nwords == 0) {
+		/* Do a handshake first. We assume that this is device-independant. */
+		memset(&ecmd, 0, sizeof(ecmd));
+		ecmd.req.cmd = ETHTOOL_GLINKSETTINGS;
+		ifr.ifr_data = (caddr_t)&ecmd;
+		rc = ioctl(sock, SIOCETHTOOL, ifr);
```
kanrajag: ioctl arg needs to be pointer to ifr (&ifr)

src/daemon/priv-linux.c
```diff
@@ -54,28 +55,88 @@ priv_open(char *file)
...
+			nwords = -ecmd.req.link_mode_masks_nwords;
+			log_debug("privsep", "glinksettings nwords is %" PRId8, nwords);
+		}
 	}
+	if (nwords > 0) {
+		memset(&ecmd, 0, sizeof(ecmd));
+		ecmd.req.cmd = ETHTOOL_GLINKSETTINGS;
+		ecmd.req.link_mode_masks_nwords = nwords;
+		ifr.ifr_data = (caddr_t)&ecmd;
+		rc = ioctl(sock, SIOCETHTOOL, ifr);
```
kanrajag: ioctl arg needs to be pointer to ifr (&ifr)

src/daemon/priv-linux.c
```diff
@@ -54,28 +55,88 @@ priv_open(char *file)
...
+		if (rc == 0) {
+			nwords = -ecmd.req.link_mode_masks_nwords;
+			log_debug("privsep", "glinksettings nwords is %" PRId8, nwords);
+		}
 	}
+	if (nwords > 0) {
+		memset(&ecmd, 0, sizeof(ecmd));
+		ecmd.req.cmd = ETHTOOL_GLINKSETTINGS;
```
kanrajag: ETHTOOL_GLINKSETTINGS cannot be issued by non-root user (yet) on our debian platform - but ETHTOOL_GSET can be (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=609226;msg=22))

did this work for you as non-root user? on which distro?

vincentbernat: I don't have a network card using `ETHTOOL_GLINKSETTINGS`, so this path is untested. I see the ability of using `ETHTOOL_GLINKSETTINGS` as non-root has been added in 4.9-rc7. I have asked if it would be possible to also push the patch to stable branches. In the meantime, I run this as root now. Could you test?

kanrajag:
lldpctl does not display MAU type advertised by port if auto-neg is disabled. maybe we should display MAU type inspite of auto-neg status?

tag_datatag(w, "current", "MAU oper type",
                            lldpctl_atom_get_str(port, lldpctl_k_port_dot3_mautype));
                        tag_end(w);

vincentbernat:
I have pushed an additional commit for that. 802.1AB mentions this is valid.

vincentbernat:
Well, this needs some adaptation of the tests. Can't do that right now. Did you get a chance to retest the GLINKSETTINGS part?

kanrajag:
Vincent. I ran some tests with the latest code. it all looks good. thanks!

vincentbernat:
❦ 24 décembre 2016 22:03 -0800, kanrajag <notifications@github.com> :

> ETHTOOL_GLINKSETTINGS cannot be issued by non-root user (yet) on our debian platform - but ETHTOOL_GSET can be
> (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=609226;msg=22))

This should be fixed in 4.9 and davem should push the appropriate commit
in stable. I'll revert the latest commit when this is released for
4.8.
-- 
When in doubt, tell the truth.
		-- Mark Twain

PR #213, opened 2016-12-03
seccomp: add mprotect to seccomp whitelist

chutz:
https://bugs.gentoo.org/600210

It seems that there are some situation where lldpd requires mprotect to run, and it is not in the seccomp whitelist.


vincentbernat:
Thanks. As for maturity, I don't enable myself seccomp in Debian because I have no idea how to cover all cases. I thought that adding integration tests would help a lot but it seems to not be enough.

PR #217, opened 2017-01-12
daemon: config netlink socket buf sizes at build time

commodo:
I admit it would be nicer to have this configurable at run-time
or better yet, introduce a mechanism for recovering from
overrun situations via polling.

We're starting to see in certain situations that on devices
with many ports (i.e. switches with around ~50 ports) lldpd
info is not always consistent.

I'm not entirely sure that the issue is with netlink
sock buffer sizes.

But we've also recentl addressed some issues for the mstpd
daemon with netlink sock buffer sizes.
So, I'm tempted to think that it's the same issue.

We [especially I] prefer to bump sock buffer sizes where needed
and not via sysctl `net.core.rmem_default` & `net.core.wmem_default`.

### DIFF
--- a/src/daemon/netlink.c
+++ b/src/daemon/netlink.c
@@ -28,6 +28,13 @@
 
 #define NETLINK_BUFFER 4096
 
+#ifndef RTNL_SND_BUFSIZE
+#define RTNL_SND_BUFSIZE	32 * 1024
+#endif
+#ifndef RTNL_RCV_BUFSIZE
+#define RTNL_RCV_BUFSIZE	256 * 1024
+#endif
+
 struct netlink_req {
 	struct nlmsghdr hdr;
 	struct rtgenmsg gen;
@@ -39,6 +46,55 @@ struct lldpd_netlink {
 	struct interfaces_device_list *devices;
 	struct interfaces_address_list *addresses;
 };
+
+
+static int
+netlink_socket_set_buffer_sizes(int s)
+{
+	int sndbuf = RTNL_SND_BUFSIZE;
+	int rcvbuf = RTNL_RCV_BUFSIZE;
+	socklen_t size = 0;
+	int got = 0;
+
+	if (setsockopt(s, SOL_SOCKET, SO_SNDBUF, &sndbuf, sizeof(sndbuf)) < 0) {
+		log_warn("netlink", "unable to set SO_SNDBUF to '%d'", sndbuf);
+		return -1;
+	}
+
+	if (setsockopt(s, SOL_SOCKET, SO_RCVBUF, &rcvbuf, sizeof(rcvbuf)) < 0) {
+		log_warn("netlink", "unable to set SO_RCVBUF to '%d'", rcvbuf);
+		return -1;
+	}
+
+	/* Now read them back from kernel.
+	 * SO_SNDBUF & SO_RCVBUF are cap-ed at sysctl `net.core.rmem_max` &
+	 * `net.core.wmem_max`. This it the easiest [probably sanest too]
+	 * to validate that our socket buffers were set properly.
+	 */
+	if (getsockopt(s, SOL_SOCKET, SO_SNDBUF, &got, &size) < 0) {
+		log_warn("netlink", "unable to get SO_SNDBUF");
+	} else {
+		if (size != sizeof(sndbuf))
+			log_warn("netlink", "size mismatch for SO_SNDBUF got '%u' "
+			         "should be '%zu'", size, sizeof(sndbuf));
+		if (got != sndbuf)
+			log_warn("netlink", "tried to set SO_SNDBUF to '%d' "
+			         "but got '%d'", sndbuf, got);
+	}
+
+	if (getsockopt(s, SOL_SOCKET, SO_RCVBUF, &got, &size) < 0) {
+		log_warn("netlink", "unable to get SO_RCVBUF");
+	} else {
+		if (size != sizeof(rcvbuf))
+			log_warn("netlink", "size mismatch for SO_RCVBUF got '%u' "
+			         "should be '%zu'", size, sizeof(rcvbuf));
+		if (got != rcvbuf)
+			log_warn("netlink", "tried to set SO_RCVBUF to '%d' "
+			         "but got '%d'", rcvbuf, got);
+	}
+
+	return 0;
+}
 
 /**
  * Connect to netlink.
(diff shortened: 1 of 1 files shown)


src/daemon/netlink.c
```diff
@@ -28,6 +28,17 @@
...
+#ifndef DEFAULT_RTNL_BUFSIZE
+#define DEFAULT_RTNL_BUFSIZE	256 * 1024
+#endif
+
+#ifndef RTNL_SND_BUFSIZE
+#define RTNL_SND_BUFSIZE	DEFAULT_RTNL_BUFSIZE
+#endif
+#ifndef RTNL_RCV_BUFSIZE
+#define RTNL_RCV_BUFSIZE	DEFAULT_RTNL_BUFSIZE
+#endif
```
vincentbernat: I would define all that directly from `configure.ac`. I can do that for you. The default is the value of `net.core.rmem_default`? So, with the value you specified, it seems to match the current default.

src/daemon/netlink.c
```diff
@@ -40,6 +51,54 @@ struct lldpd_netlink {
 	struct interfaces_address_list *addresses;
 };
 
+
+static int
+netlink_socket_set_buffer_sizes(int s)
+{
+	int sndbuf = RTNL_SND_BUFSIZE;
+	int rcvbuf = RTNL_RCV_BUFSIZE;
+	int read = 0, size = 0;
```
vincentbernat: You shadow the `read()` function. I suggest `got`.

src/daemon/netlink.c
```diff
@@ -40,6 +51,54 @@ struct lldpd_netlink {
...
+
+	if (setsockopt(s, SOL_SOCKET, SO_SNDBUF, &sndbuf, sizeof(sndbuf)) < 0) {
+		log_warn("netlink", "unable to set SO_SNDBUF to '%d'");
+		return -1;
+	}
+
+	if (setsockopt(s, SOL_SOCKET, SO_RCVBUF, &rcvbuf, sizeof(rcvbuf)) < 0) {
+		log_warn("netlink", "unable to set SO_RCVBUF to '%d'");
+		return -1;
+	}
```
vincentbernat: What's important is the receive buffer, right? We never send anything I think. This would half the memory usage needed. See my general comment about this.

src/daemon/netlink.c
```diff
@@ -40,6 +51,54 @@ struct lldpd_netlink {
...
+	 * SO_SNDBUF & SO_RCVBUF are cap-ed at sysctl `net.core.rmem_max` &
+	 * `net.core.wmem_max`. This it the easiest [probably sanest too]
+	 * to validate that our socket buffers were set properly.
+	 */
+	if (getsockopt(s, SOL_SOCKET, SO_SNDBUF, &read, &size) < 0) {
+		log_warn("netlink", "unable to get SO_SNDBUF");
+	} else {
+		if (size != sizeof(sndbuf))
+			log_warn("netlink", "size mismatch for SO_SNDBUF got '%d' "
+			         "should be '%d'", size, sizeof(sndbuf));
```
vincentbernat: `sizeof()` is `%zu`.

src/daemon/netlink.c
commodo: ack

src/daemon/netlink.c
commodo: ack

commodo:
regarding your question:
```
Just to be sure, you have warnings in your logs with unable to receive netlink answer: no buffer space available?
```

i only got some report [in a customer setup] that a module [which is using logic from lldpd by querying for TLVs], is not functioning correctly ;
when i started investigating this, there wasn't much in the log to help, since it's on LEDE/OpenWrt systems, logs are circular and they get overwritten after a while

so, based on the `mstpd` experience we've had [with the same device class ; i.e. ~50 port switch], I'm trying to address this from the netlink perspective, where if too many ports are being brought up [link events], the netlink buffers seem insufficient

we're trying to address some of these netlink cases [each one being treated individually], as we've started to see netlink-buffer-overrun issues in various other modules [ some internal ]

and this is the lldpd case :)
[ so, in short, i'm trying to get support off my back :p ]

commodo:
on LEDE/OpenWrt  `net.core.rmem_max ` & `wmem_max` are around 160k ; but we've bumped those up with `sysctl.conf`

src/daemon/netlink.c
commodo: actually, i chose this default value arbitrarily ; 
to fit general needs [ of ours ] ; it could be adapted to something else

in the kernel source, the default value of `rmem_default` is
```
#define SKB_TRUESIZE(X) ((X) +                      \
             SKB_DATA_ALIGN(sizeof(struct sk_buff)) +   \
             SKB_DATA_ALIGN(sizeof(struct skb_shared_info)))

#define _SK_MEM_OVERHEAD        SKB_TRUESIZE(256)
#define _SK_MEM_PACKETS         256

#define SK_RMEM_MAX             (_SK_MEM_OVERHEAD * _SK_MEM_PACKETS)

__u32 sysctl_rmem_default __read_mostly = SK_RMEM_MAX;
```

on our system it shows up as ~160k ; 

on my dev machine:
```
net.core.rmem_default = 212992
net.core.wmem_default = 212992
```

i guess it can vary from system to system, depending how the kernel is compiled

src/daemon/netlink.c
commodo: yep ; the recv buffer is a bit more important ;

commodo: i think 32k might also be safe to add for SO_SNDBUF ; if we're not sending anything

commodo:
re-spinned ; i hope i got everything correctly ; 
apologies if i missed anything ; 
i've been multi-tasking a bit too much lately

commodo:
and i'll take you up on the offer to convert to `configure.ac` ; if i had a bit more energy i would have attempted to do it 
sorry about that :)

commodo:
sorry about the multiple re-spins ; apparently the compiler options on the OpenWrt build that I'm using are a bit too permissve

commodo:
only one build failed this time;
https://travis-ci.org/vincentbernat/lldpd/jobs/191348122
looks unrelated to my changes

vincentbernat:
Yep, Travis is a bit unreliable. I'll merge your changes tomorrow or this week-end.

PR #218, opened 2017-01-14
Auto-adapt Netlink socket size

vincentbernat:
When netlink socket receive buffer is too small, increase it automatically up to a compile-time limit.

@commodo, there are some changes compared to your PR:
 - use of `<` instead of `!=` as the kernel seems to round the result
 - by default, we don't change the buffer size
 - removed a check about the size of the int returned by the kernel (don't think that's useful)

I have tried with very small buffers and it seems to work.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,8 @@ lldpd (0.9.6)
     + Support for newer ethtool interface on Linux
       (ETHTOOL_GLINKSETTINGS) and additional speed settings.
     + Current MAU type is displayed even when autoneg is off.
+    + Increase netlink receive buffer by default. Can be changed at
+      compile-time through ./configure.
   * Fix:
     + Correctly parse LLDP-MED civic address when the length of the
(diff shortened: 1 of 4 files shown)


commodo:
Hey

Atm I'm traveling.
I'm in Saas Fe :)

Will take a look closer on Monday.

Overall looks good.
I would compress some parts ; mostly the one where you're changing the
buffer size to 4 then 1 MB, and just set it to 1 from the start.

vincentbernat:
I have started some refactoring to minimize the duplicate code, but I am pulled by something else. I'll finish tomorrow. Hold on.

commodo:
hey ; so, from my side, code looks good

for my current needs, I've backported a patch that just bumps socket buffer sizes to some arbitrary sizes that will avoid the issue

then at a point in time, we will upgrade to a official version that supports this feature
[ it's still undecided ; we may backport the entire changset here ]

thanks
Alex

PR #224, opened 2017-03-17
Increase number of lldp neighbors to 32.

SanderVermin:
In our company we have multiple situation where a number of devices are connected to a un-managed switch. So this gives a number higher number of lldp neighbors in the remote table.


vincentbernat:
Thanks!

PR #228, opened 2017-04-12
Issue153

cbrasho:
this is in response to https://github.com/vincentbernat/lldpd/issues/153
not sure if calling lldpd_cleanup(cfg) in daemon/client.c is enough to reset the configuration.


src/client/lldpcli.c
```diff
@@ -438,7 +474,6 @@ main(int argc, char *argv[])
 	ctlname = lldpctl_get_default_transport();
 
 	signal(SIGHUP, SIG_IGN);
-
```
commodo: nitpick: extra whitespace

src/daemon/client.c
```diff
@@ -110,6 +110,19 @@ client_handle_set_configuration(struct lldpd *cfg, enum hmsg_type *type,
 		cfg->g_config.c_paused = config->c_paused;
 		levent_send_now(cfg);
 	}
+	/* reset */
+	if (CHANGED(c_reset) &&
+		config->c_reset == 1) {
+		log_debug("rpc", "client asked to reset lldpd's config");
+		cfg->g_config.c_reset = 1;	//mark that reset is being processed by daemon
+									//so that client won't accept any new reset
```
vincentbernat: You don't really need an "in progress" marker. lldpd is totally "evented", so nothing will execute in parallel until the end of this function. So, nobody will be able to observer a case where `cfg->g_onfig.c_reset == 1`. You can remove that and some other code on the client side about that.

vincentbernat:
Also needed (but I can do it after merging):

 - update the manual page with the new command
 - add an integration test

src/client/lldpcli.c
```diff
@@ -192,6 +192,39 @@ cmd_resume(struct lldpctl_conn_t *conn, struct writer *w,
...
+    lldpctl_atom_t *config = lldpctl_get_configuration(conn);
+    if (config == NULL) {
+        log_warnx("lldpctl", "unable to get configuration from lldpd. %s",
+            lldpctl_last_strerror(conn));
+        return 0;
+    }
+    if (lldpctl_atom_get_int(config, lldpctl_k_config_reset) == 1) {
+        log_debug("lldpctl", "lldpd  already has a reset config request in it's queue. Please try later.");
+        lldpctl_atom_dec_ref(config);
+        return 1;
```
vincentbernat: This is an impossible case. No client can never get a value of 1 at this point.

src/lib/lldpctl.h
```diff
@@ -793,6 +793,7 @@ typedef enum {
 	lldpctl_k_custom_tlv_oui_subtype,	/**< `(I,WO)` custom TLV subtype. Default is 0 (1 byte) */
 	lldpctl_k_custom_tlv_oui_info_string,	/**< `(I,WO)` custom TLV Organizationally Unique Identifier Information String (up to 507 bytes) */
 	lldpctl_k_custom_tlv_op,		/**< `(I,WO)` custom TLV operation */
+	lldpctl_k_config_reset,	/**< `(I)` lldpd is reset */
```
vincentbernat: This is `(I,WO)`. Also, small nitpick, uses some tabs to align it with the other comments.

src/client/lldpcli.c
```diff
@@ -192,6 +192,39 @@ cmd_resume(struct lldpctl_conn_t *conn, struct writer *w,
...
+        log_warnx("lldpctl", "unable to ask lldpd to reset operations. %s",
+            lldpctl_last_strerror(conn));
+        lldpctl_atom_dec_ref(config);
+        return 0;
+    }
+    log_info("lldpctl", "lldpd should reset configs");
+    lldpctl_atom_dec_ref(config);
+    return 1;
+}
+
```
vincentbernat: This whole function should use tabs instead of spaces for indentation.

cbrasho:
thanks for your feedback.
I will get back with a new approach ASAP (could take some time since I have
other engagements right now).

commodo:
So, an update from my side.

This would probably need to be picked up by someone else; and probably re-designed.
What happened is: I left Riverbed around December 2017.
Up until then, I tried to delegate this task [since I had my hands full with other work].
@cbrasho also left Riverbed some-time after I did.

I don't know if there is direct interest [within Riverbed] to continue this. Probably other stuff is more important.

This can be left as-is [from my side] as a reference-point.
Or, if it's superseded by something else, this can be closed.
I probably should have mentioned this stuff earlier, but forgot.
Github doesn't show too much recent activity [I guess].

Thanks
Alex

vincentbernat:
We can keep it open. I don't mind having opened PR. This is easier to find than a closed PR. I have also very little time and have some other issues to handle before this. So, let's not make promises here!

PR #230, opened 2017-05-19
fix -Wimplicit-fallthrough found by gcc7

alagoutte:
i think, there is a "real" (minor) bug only for interfaces-linux.c


vincentbernat:
Oh, great! Thanks!

PR #233, opened 2017-07-11
lldp: clear station bit if any other capability is enabled

steven-webster:
The 801.1AB spec states that the station only capability should
not be set in conjunction with any other (capability) bits.

It is possible for the lldpd interface update routine to set the
station bit, while the lldpd chassis update routine may, at a later
time, set the router capability bit. In this case the station bit
is not cleared.

This commit clears the station bit if any other capability is set.


vincentbernat:
Thanks!

PR #238, opened 2017-08-27
add option to override default lldpcli config file locations

tannhaeuser:
This pr adds a cmd line option to override the default config files/directory locations (e.g. /etc/lldpd.conf and /etc/lldpd.d) used by lldpd to trigger configuration changes executed by lldpcli on startup.

This option is very helpful if one would like to run more than one instance of lldpd on a box (e.g. I use network namespaces to simulate a large network and I would like to avoid the usage of chroot, network namespaces or the need of running lldpcli manually for each instance) . It also supports automated tests using different lldpd configurations as it avoids the hassle of copying config files.

### DIFF
--- a/src/daemon/lldpd.8.in
+++ b/src/daemon/lldpd.8.in
@@ -34,6 +34,7 @@
 .Op Fl M Ar class
 .Op Fl H Ar hide
 .Op Fl L Ar lldpcli
+.Op Fl O Ar configfile
 .Sh DESCRIPTION
 .Nm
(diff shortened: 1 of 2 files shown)


vincentbernat:
This seems good. There are some indent problems. I can fix them for you if you want.

tannhaeuser:
thank you for the fast response.

no need to support my ignorance :) hope e6e11a0 fixes all issues.

vincentbernat:
I have squashed through Github. Thanks!

PR #240, opened 2017-08-29
client: add show interfaces command

steven-webster:
There are several lldpcli configure commands that take port name(s)
as an optional parameter.

Currently, it is not possible to verify interface specific configuration
for lldpd without access to a neighbour switch/host or packet capture.

It would be useful to have a show interfaces command, similar to the
show neighbors command that allows a user/agent to query the local
TLV information.

This commit leverages the existing port display infrastructure to
enable display of local interfaces.  There are a few differences:

- Do not display the chassis RID, as the local index is always 0
- Show TTL as a seperate TLV based on tx_hold, tx_interval config,
  as internally the port TTL atom info is 0, and the chassis TTL
  atom info has been depreciated.

### DIFF
--- a/src/client/display.c
+++ b/src/client/display.c
@@ -349,7 +349,8 @@ display_port(struct writer *w, lldpctl_atom_t *port, int details)
 
 	tag_datatag(w, "descr", "PortDescr",
 	    lldpctl_atom_get_str(port, lldpctl_k_port_descr));
-	if (details)
+	if (details &&
+	    lldpctl_atom_get_int(port, lldpctl_k_port_ttl) > 0)
 		tag_datatag(w, "ttl", "TTL",
 		    lldpctl_atom_get_str(port, lldpctl_k_port_ttl));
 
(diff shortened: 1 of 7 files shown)


vincentbernat:
Wow, there are even tests! At first look, it seems good. I'll try to merge before the end of the week.

src/client/display.c
```diff
@@ -675,6 +714,54 @@ display_interfaces(lldpctl_conn_t *conn, struct writer *w,
...
+		for (lldpctl_map_t *protocol_map =
+			 lldpctl_key_get_map(lldpctl_k_port_protocol);
+		     protocol_map->string;
+		     protocol_map++) {
+			if (!strcasecmp(proto_str, protocol_map->string)) {
+				protocol = protocol_map->value;
+				break;
+			}
+		}
+	}
```
vincentbernat: This code seems to be unreachable: there is no way to specify a protocol for a local interface.

src/client/display.c
```diff
@@ -474,6 +475,36 @@ display_port(struct writer *w, lldpctl_atom_t *port, int details)
...
+		log_warnx("lldpctl", "not able to get configuration. %s",
+		    lldpctl_last_strerror(conn));
+		return;
+	}
+
+	tx_hold = lldpctl_atom_get_int(configuration, lldpctl_k_config_tx_hold);
+	tx_interval = lldpctl_atom_get_int(configuration, lldpctl_k_config_tx_interval);
+
+	if (asprintf(&ttl, "%lu", tx_hold*tx_interval) == -1) {
+		log_warnx("lldpctl", "not enough memory to build TTL.");
```
vincentbernat: It's unlikely we ever hit this path, but it should have a goto to `lldpctl_atom_dec_ref()` at the end of the function or we will crash when using `ttl`.

steven-webster: Oh dear ... good catch.

src/client/display.c
steven-webster: Right.  This can be removed, as if I understand correctly, the protocol used by a local interface is determined by what it receives.

vincentbernat:
Thanks!

PR #247, opened 2017-09-30
daemon: don't use SIGSTOP when using Upstart

### LINKED ISSUE #245: After lldpd got SIGKILL, respawn didn't work
# Bug description
I realized the issue when I restarted the lldpd several times within short time period. Then I found lldpd failed to respawn and kept the state:

status lldpd
lldpd start/running

After I looked at the log, I found that there was a slow shut down which 5 seconds after we sent the SIGTERM to lldpd service, if still running, upstart job will send SIGKILL to lldpd. But after that the lldpd didn't respawned by upstart.(There is no limit for respawn).

lldpd version is 0.9.4

## Steps to reproduce the problem

"killall -s 9 lldpd" can reproduce the problem:

        init: lldpd main process (21023) killed by KILL signal
	init: lldpd main process ended, respawning
	lldpd[22027]: unable to create control socket because it already exists
	lldpd[22027]: check if another instance is running
	lldpd[22027]: another instance is running, please stop it
	lldpd[22027]: giving up
	init: lldpd main process (22027) terminated with status 1
	init: lldpd main process ended, respawning

After this, there is no log shows that it respawned.

However, "restart lldpd" can recover from this state:

        lldpd[21023]: unable to create control socket because it already exists
        lldpd[21023]: check if another instance is running
        lldpd[21023]: unable to connect to socket /var/run/lldpd.socket: Connection refused
        lldpd[21023]: old control socket is present, clean it

vincentbernat:
This is buggy because if lldpd happens to stop before sending SIGSTOP,
Upstart won't respawn it. See:
 https://bugs.launchpad.net/upstart/+bug/712351

Instead, let Upstart detects the double fork. This should work just
fine as there is no "trap" before daemonizing.

Fix #245

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,7 @@ lldpd (0.9.9)
       each of them ("show interfaces").
   * Fix:
     + Don't remove interfaces when they are released from a bridge.
+    + Don't use "expect stop" with Upstart. It's buggy.
 
 lldpd (0.9.8)
   * Changes:
--- a/debian/lldpd.upstart
+++ b/debian/lldpd.upstart
@@ -9,7 +9,7 @@ description "LLDP daemon"
 start on net-device-up IFACE=lo
 stop on runlevel [06]
 
-expect stop
+expect fork
 respawn
 
 script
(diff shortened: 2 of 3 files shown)


vincentbernat:
Unfortunately, the exact same bug exists when using `expect fork`.

PR #251, opened 2017-10-20
priv: correctly handle lldpcli exit

### LINKED ISSUE #250: lldpcli zombie
When configured without privilege separation, starting `lldpd` creates a `lldpcli` zombie -- the daemon never calls waitpid() for the process forked by lldpd_configure().

This does not occur with privilege separation because sig_chld() in src/daemon/priv.c happens to call waitpid(-1, ...) if a SIGCHLD is received and the monitored child is still running.

I am attaching a quick hack to get rid of the zombie; it sets SIGCHLD to SIG_IGN, so terminated children will not become zombies. This may not be entirely clean, but it works for me ...

Note that there are some conditions for this patch:
* We don't want to ignore SIGCHLD before lldpd_get_lsb_release() since the latter uses waitpid().
* We assume that nobody will call wait() or waitpid() afterwards, as this may block forever.
* We rely on `ldpcli`, if running, not to exit before we enter levent_loop(); it sends at least a "resume" command and waits for the response (even if /etc/lldpd.conf is missing or empty).

[lldpcli-zombie.patch](https://github.com/vincentbernat/lldpd/files/1400835/lldpcli-zombie.patch.txt)

vincentbernat:
When lldpcli exits, we must acknowledge its death with
`waitpid()`. There were two missing cases:

 - when lldpcli exits before setting the SIG_CHLD signal
 - when privilege separation was not configured

For the first case, we call the function associated to the signal
ourselves after the signal is configured. This function is idempotent.

For the second case, we setup the SIG_CHLD signal even when privilege
separation is disabled.

Fix #250

### DIFF
--- a/src/daemon/priv.c
+++ b/src/daemon/priv.c
@@ -488,7 +488,9 @@ sig_pass_to_chld(int sig)
 	errno = oerrno;
 }
 
-/* if parent gets a SIGCHLD, it will exit */
+/* If priv parent gets a SIGCHLD, it will exit if this is the monitored
+ * process. Other processes (including lldpcli)) are just reaped without
+ * consequences. */
 static void
 sig_chld(int sig)
 {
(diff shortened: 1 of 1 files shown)


tbetker:
635724519137dde5adc749f6320f225373fd63d5 + b274c9edafad09a71509d54a5ad8b4ca487790ba look good to me. However, I would suggest keeping the original `#ifdef ENABLE_PRIVSEP` version of sig_chld() as it is, and just add another `#else` version which does `while (waitpid(-1, NULL, WNOHANG) > 0);`, minimizing code changes.

Anway, I will try to test these two patches on Monday.

vincentbernat:
I have amended the second patch with your suggestion. Tests seem OK.

tbetker:
I don't have the amended patch, but I have tested 6357245 + b274c9e today, and it worked.

vincentbernat:
FI, you can also get the patch by appending `.patch` to the PR URL: https://patch-diff.githubusercontent.com/raw/vincentbernat/lldpd/pull/251.patch.

PR #253, opened 2017-11-08
lldpcli: document json0 output format

gregoryp:
Hello,

json0 format has been added but not documented in lldpdcli/lldpdctl help output. PR fixes just that.

Thank you!


vincentbernat:
Thanks!

PR #256, opened 2017-11-19
dot3: as PD device, echo back PSE allocated value

### LINKED ISSUE #243: TLV send from PD does not show new allocated power for PSE
I have lldp working on an embedded device using Android. Everything appears to work correctly. However, I went for testing at a lab prior to getting certified and they said there is an error in the protocol. I send a message the says I have 13 watts and request 24.6 watts. The PSE send back a message that says I requested 24.6 watts and it allocated 24.6. The next message I send has the original values, 13 and 24.6.
The lab says I should be sending a message with 24.6 allocated and 24.6 requested.

My conf file looks like this:
configure system hostname "Company"
configure system description "Company Device"
configure system interface pattern eth0
configure dot3 power pd supported enabled powerpairs signal class class-4 type 2 source pse priority high requested 24600 allocated 13000

### LINKED ISSUE #249: 10 second requirement for Power via MDI TLV
From 802.3at Clause 33.6.2: 
Under normal operation, an LLDPDU containing a Power via MDI TLV with an updated value for "PSE allocated power value" field shall be sent within 10 seconds of receipt of a LLDPDU containing a Power via MDI TLV where the "PD requested power value" is different from the previously communicated value.

Testing showed various times up to about 25 seconds. I believe we were waiting for the 30 second timer. I changed the timer to 10 seconds to verify this and met the timing requirement. However I didn't want to run the loop that often. I put a call to "levent_send_now" at the end of the LLDP_TLV_DOT3_POWER case in lldp_decode and now send the PDU back to the PSE instantly.

Mike

vincentbernat:
Dot3 power TLV contains an allocated value and a requested value. When
PSE allocates some power and says so in its TLV, PD device is expected
to echo back (within 10 seconds) the received value in its own TLV. We
handle this part automatically.

Fix #243
Fix #249

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,7 @@ lldpd (0.9.9)
   * Changes:
     + lldpcli can now display local interfaces with LLDP data sent on
       each of them ("show interfaces").
+    + As Dot3 PD device, echo back allocated value from PSE device.
   * Fix:
     + Don't remove interfaces when they are released from a bridge.
     + Don't use "expect stop" with Upstart. It's buggy.
--- a/src/daemon/lldpd.c
+++ b/src/daemon/lldpd.c
@@ -944,6 +944,37 @@ lldpd_hide_all(struct lldpd *cfg)
 	}
 }
 
+/* If PD device and PSE allocated power, echo back this change. If we have
+ * several LLDP neighbors, we use the latest updated. */
+static void
+lldpd_dot3_power_pd_pse(struct lldpd_hardware *hardware)
+{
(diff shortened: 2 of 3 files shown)


vincentbernat:
@mikem388, could you test if this works for you? Or at least, check the test logic in https://github.com/vincentbernat/lldpd/pull/256/files#diff-26675baf21cdf796c90b2c4d3a4dfae0R60.

mikem388:
Vincent,

The logic looks correct. 

Thanks for your efforts,
Mike
________________________________________

PR #259, opened 2017-12-07
LICENSE: add title and copyright notice

waldyrious:
The title is not legally mandated, but it's recommended in the license template text (see http://choosealicense.com/licenses/isc/ and https://opensource.org/licenses/isc-license).

The copyright notice is required, IIRC.

*This PR is part of a project to improve the consistency and visibility of the ISC license. See https://github.com/github/choosealicense.com/issues/377 for more details.*

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,7 @@
+The license below applies to most, but not all content in this project.
+Files with different licensing and authorship terms are marked as such.
+That information must be considered when ensuring licensing compliance.
+
 ISC License
 
 Copyright (c) 2008-2017, Vincent Bernat <vincent@bernat.im>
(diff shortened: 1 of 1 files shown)


vincentbernat:
Unfortunately, this is an incorrect summary. Each file has its own copyright notice that people should look at if they need to. The LICENSE file only contains the license for the majority of the code (and I only put it here for GitHub to detect it correctly). Even when the code is ISC licensed, I may not be the only copyright holder.

waldyrious:
I see. Should I add a disclaimer above the license text explaining that it applies to most, but not all the contents of the repo, and that individual file headers should be checked regardless?

vincentbernat:
Yes, it would be OK then.

vincentbernat:
Thanks!

Issue #244, opened 2017-09-20
Can Chassis ID's value  be locally assigned ?  It seems to be Mac address or system name now. 


vic-lin-accton:
Hi Sir,

   I can not find any instruction to modify Chassis ID using our own local use byte string.
Check those Usage doc, it seems to us only use MAC to compute or system name
for Chassis ID.

- Vic

vincentbernat:
Yes, this is currently not possible. I'll try to add this possibility in the next month.

vincentbernat:
Fixed by #263.

PR #263, opened 2017-12-28
Check if there is user defined string before use chassis name for cha…

vic-lin-accton:
Hi Vincent,
    This is for issue #244 opened 20 Sep 2017. Please review it.
Thanks
-Vic

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,7 @@
 lldpd (1.0.0)
+  * Changes:
+    + Chassis ID can be set to an arbitrary value with "configure system
+      chassisid".
   * Fix:
     + Ensure chassis-related changes are propagated immediately.
 
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,7 @@
+lldpd (1.0.0)
+  * Fix:
+    + Ensure chassis-related changes are propagated immediately.
+
 lldpd (0.9.9)
   * Changes:
     + lldpcli can now display local interfaces with LLDP data sent on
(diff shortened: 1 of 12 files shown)


vincentbernat:
Thanks! This looks good. However, I would like to not add additional command-line options. This should be configurable through lldpcli only. I can fix that this week-end if you want.

vic-lin-accton:
Yes, please fix it. Thanks

vic-lin-accton:
It's not urgent for me. You can fix that when you have time. Thanks

vincentbernat:
I have edited your first commit for small blank changes (space to tab). Other commits enable to use lldpcli to push the change. Let's see if tests go through.

vic-lin-accton:
Great. Thanks

vic-lin-accton:
Override seems to be more correct wording than user defined string for user to configure the chassis id, is it ?

vincentbernat:
Where exactly should "override" be used?

vic-lin-accton:
Sorry, it is in the output of lldpcli  "show config" as below. The code should be in display_configuration()/display.c . I think "description", "platform" are examples, too.

Configuration:
  Transmit delay: 30
........
  Interface pattern for chassis ID: !*
  User defined string for chassis ID: (none)
  Override description with: 
  Override platform with: Linux
  Override system name with: 
  Advertise version: yes

vincentbernat:
I have changed the string to "Override chassis ID with".

PR #271, opened 2018-02-26
Fix NULL pointer dereference in lldpd_dot3_power_pd_pse

axjowa:
selected_port might be NULL after the loop if, for example,
no PSE was found among the neighbours.

This fixes a crash that occured with a device connected to
a non-PoE, LLDP-enabled switch via a PoE midspan.


vincentbernat:
Thanks!

PR #272, opened 2018-03-12
handle lldpStatsRemTablesLastChangeTime correctly when items are removed

jonasj76:
When a port is removed, the time has to be updated. The last removal time is registered per local port, and this timestamp will be used as lldpStatsRemTablesLastChangeTime if it is the latest timestamp.
Also, the lldpStatsRemTablesDeletes is  always increased when an entry in the table is deleted.


vincentbernat:
Thanks!

PR #273, opened 2018-03-12
client: add alternative way to configure port description

jonasj76:
The existing port description CLI command only allows user configured description if setting port id subtype to local. This patch introduces an subtype independent way to configure port description.


vincentbernat:
Thanks! Could you also update the manual page for lldpcli?

jonasj76:
Manual page updated.

Should the `description` setting in `configure lldp portidsubtype local <value> description <description>` be kept?

PR #276, opened 2018-03-28
Install headers in pkgincludedir in addition to includdir

rschaefer2:
I have a use case where I'd like to include lldpd headers, but not everything else in /usr/include. This will install the headers in their own package directory in addition to includedir. Original location is left for backwards compatibility for packages including the headers from the original location.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -4,6 +4,7 @@ AM_LDFLAGS = $(LLDP_LDFLAGS)
 
 noinst_LTLIBRARIES = libcommon-daemon-lib.la libcommon-daemon-client.la
 include_HEADERS    = lldp-const.h
+pkginclude_HEADERS    = lldp-const.h
 
 libcommon_daemon_lib_la_SOURCES = \
 	log.c log.h version.c \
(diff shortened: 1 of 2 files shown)


vincentbernat:
Thanks for the patch but I don't quite understand. What's "everything else"? Only these two files should be installed. Installing the same files at two different locations would just be confusing.

rschaefer2:
/usr/include/ is the path for system headers. Package headers ideally should be separate from the system headers. If I have a program and would like to include a different set of headers than those in /usr/include/, but also include the lldpd.h headers, I am in somewhat of a pickle since the lldpd headers are installaed alongside the system headers.

This may seem like a weird use case. I am running into this problem when using clang-tidy with a program that is cross-compiled with a different sysroot. clang-tidy uses it's own system headers. If I attempt to include the sysroots /usr/include, the headers will conflict with the system headers clang-tidy uses and cause errors. Since this is also the location of the lldpd headers, I am unable to include the lldpd headers with clang-tidy.

vincentbernat:
What about using `./configure --includedir=/usr/include/lldpd`?

rschaefer2:
Ah, much cleaner. I don't know how I missed that option. Thanks!

PR #285, opened 2018-06-12
priv: drop most privileges in monitor, only keep CAP_NET_RAW/ADMIN

vincentbernat:
On Linux, we mostly rely on CAP_NET_RAW. Only keep that one. However,
we also write to ifalias, which needs CAP_NET_ADMIN. We could let user
choose at runtime if they want to grant this capability or not.
Currently, a user can turn it on/off at any time.

Access to SNMP socket may also be problematic. We need some solid
solution about that before merging.

Is it safe to use the same UID for the monitored and the unprivileged
process? Signals are mostly harmless. As for ptrace, since the
monitored process as more capabilities, this will not be allowed by
Linux.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,8 +2,8 @@ lldpd (1.0.2)
   * Changes:
     + On Linux, the monitor process will now drop its privileges
       instead of running as root. It will keep CAP_NET_RAW and
-      CAP_NET_ADMIN capabilities. When using SNMP AgentX feature, the
-      access to the socket may require to grant access to _lldpd user.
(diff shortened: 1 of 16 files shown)


vincentbernat:
Remaining items:

 - If `-x` option is provided, we could add `CAP_FOWNER` to not have any problem with AgentX support. We could check the permissions of the AgentX socket if it exists (maybe too random).
 - Compile-time option for capabilities to keep? Some users may want to restrict to `CAP_NET_RAW`.
 - On Ubuntu Trusty (and therefore on Travis), `libcap-dev` is missing the `.pc` file.

PR #286, opened 2018-06-13
Read all notifications in lldpctl_recv.

GustavWi:
Can otherwise lead to unbounded growth in input_buffer if
lldp devices send notifications simultaneously thus
a socket callback contains multiple notifications
and only the first one is cleared. This leads to continous
growth of the input buffer and will crash the system.


vincentbernat:
Thanks!

PR #288, opened 2018-06-21
Add links to the IEEE Std 802.1AB revisions

jpoliv:


vincentbernat:
Thanks for the patch! lldpd only implements 802.1AB-2005 (like any other LLDP implementations AFAIK). I think it is misleading to add links to more recent revisions.

vincentbernat:
Closing this PR as we don't implement more recent versions of 802.1AB.

PR #289, opened 2018-06-21
Add support for PD PoE negotiation.

GustavWi:
There are three TLVs for CDPv2 PoE negotiation.
Power Consumption: Current maximum power consumption of PD.
Power Request: Wanted maximum power consumption of PD.
Power Available: Power output from PSE.

Only used if lldp PoE is not supported by switch.

A cisco switch which does support both lldp and cdp will
use the protocol which is first to transmit a package.


src/daemon/lldpd.c
```diff
@@ -986,8 +986,13 @@ lldpd_dot3_power_pd_pse(struct lldpd_hardware *hardware)
...
 		if (port->p_hidden_in)
 			continue;
-		if (port->p_protocol != LLDPD_MODE_LLDP)
+#ifdef ENABLE_CDP
+		if (port->p_protocol != LLDPD_MODE_LLDP && port->p_protocol != LLDPD_MODE_CDPV2)
+			continue;
+#else
+		if (port->p_protocol != LLDPD_MODE_LLDP
 			continue;
+#endif
```
vincentbernat: You don't need this `#ifdef`. If `ENABLE_CDP` is not defined, `port->p_protocol` cannot be equal to `LLDPD_MODE_CDPV2`, so no issue in having a single condition.

src/daemon/protocols/cdp.c
```diff
@@ -201,7 +201,38 @@ cdp_send(struct lldpd *global,
...
+	if ((version >= 2) &&
+	    (port->p_power.powertype != LLDP_DOT3_POWER_8023AT_OFF) &&
+	    (port->p_power.devicetype == LLDP_DOT3_POWER_PD) &&
+	    (port->p_power.requested > 0) &&
+	    (port->p_power.requested <= 655)) {
+		if (port->p_power.requested != port->p_power.allocated) {
+			port->p_cdp_power.request_id++;
+			log_debug("cdp", "requested: %d, allocated:%d", port->p_power.requested, port->p_power.allocated);
+		}
+		u_int16_t consumption = port->p_power.allocated ? port->p_power.allocated : 154;
```
vincentbernat: What is the magic value 154? Maybe use a constant?

src/daemon/protocols/cdp.c
```diff
@@ -522,6 +559,21 @@ cdp_decode(struct lldpd *cfg, char *frame, int s,
 			TAILQ_INSERT_TAIL(&port->p_vlans,
 					  vlan, v_entries);
 			break;
+#endif
+#ifdef ENABLE_DOT3
+		case CDP_TLV_POWER_AVAILABLE:
+			CHECK_TLV_SIZE(12, "Power Available");
```
vincentbernat: You request 12 bytes, but only read 8 bytes. Is that correct?

vincentbernat:
Thanks for the contribution! Other than the small nitpicks commented online, the change looks good.

src/daemon/protocols/cdp.c
GustavWi: There is a TLV named Power Management Level which seems to be the value the switch wants the PD to display in its Power Consumption TLV. I have not added support for it because it seems unnecessary.
https://www.cisco.com/assets/sol/sb/SG220_Emulators/SG220_Emulator_v1-0-0-18_20140626/help/Discovery17.html

GustavWi:
I realised that the CDP protocol uses the Power values at the PSE and not the PD which is the case for lldp. I have a cisco switch and when I requested 25.5W via CDP I got 25.5W at the switch port. Then I tried lldp and asked for 25.5W and got 30W at the Switch output.

src/daemon/protocols/cdp.c
GustavWi: Done

src/daemon/lldpd.c
GustavWi: Done

vincentbernat:
I don't know enough how PoE works to understand the difference. Is it something you need to fix or is it good enough to merge? From the code point of view, I am OK to merge.

GustavWi:
I have one problem:
On a cisco switch which PoE negotiates via CDP to the device. If I then enable LLDP the cisco switch will broadcast lldp frames with default Power TLVS. It won't use it for power negotiations but will still broadcast intial values in this TLV like Power Allocated (13W) and Power Requested (13W). lldpd will then read this broadcast packet and assume 13W has been allocated and override CDP. Then lldpd will toggle between 25.5W from CDP and 13W from lldp. In this case the lldp packets should not be used as PoE negotiation. It works in the reverse becaue I ignore CDP packets with initial values because the request-id=0 in the TLV so only LLDP is used for PoE.

Do you have any suggestion for a solution to this problem?

GustavWi:
Yes the software is from april 2018.

dennypage:
Perhaps report a bug to Cisco?

vincentbernat:
I would be relunctant to implement a workaround directly into lldpd as this would be a can of worms, but it would be easy to fix that in your own fork by looking if there is a CDPv2 neighbor before "answering" to LLDP PoE requests.

GustavWi:
I added a commit to handle the lldp generic broadcast while cdp is used for PoE:
https://github.com/GustavWi/lldpd/tree/fix-lldp-spam-when-cdp-used

src/daemon/protocols/cdp.c
```diff
@@ -522,6 +583,29 @@ cdp_decode(struct lldpd *cfg, char *frame, int s,
...
+			/* check if it is a respone to a request id */
+			if (PEEK_UINT16 > 0) {
+				port->p_cdp_power.management_id = PEEK_UINT16;
+				port->p_power.allocated = PEEK_UINT32;
+				port->p_power.allocated /= 100;
+				port->p_power.supported = 1;
```
GustavWi: I should perhaps remove p_power.support and p_power.enabled since it is incorrect and irrelevant?

vincentbernat:
About the second commit, I am quite undecided. I see several possibilities for this code to go wrong. Cisco could fix the bug, the equipment could be transferred to another switch speaking only  LLDP and the new power stuff would be ignored. The equipment could be on an LLDP switch but receiving CDP frames from a neighbor CDP-aware switch, disabling LLDP power handling.

As long as the code is simple, we can do power negotiation directly inside lldpd. But if there is a need for policy, I think it should be delegated to an outside process (or handled by a local patch in case like yours).

PR #294, opened 2018-07-18
Add support to disable MED policy from port.

maloy-ghosh:
Before it was possible to enable/modify network policy TLVs on a port
but it was not possible to disable them[1].

This commit introduces a new command in the form `configure ports <port>
med policy application <app> disabled` to explicitly disable an
application TLV from a port.

[1] https://github.com/vincentbernat/lldpd/issues/292


vincentbernat:
I am sorry, I did miss your PR for some reason and only go through it now. If you still remember, could you use `unconfigure` instead of appending a `disabled`?

PR #311, opened 2018-12-07
daemon: create runtime directories with systemd-tmpfiles

eworm-de:


configure.ac
```diff
@@ -346,9 +353,9 @@ dnl On autoconf 2.69 and before, runstatedir is not configurable, let be able to
 if test "x$runstatedir" = x; then
     AC_SUBST([runstatedir], ['${localstatedir}/run'])
 fi
-lldp_ARG_WITH([privsep-chroot], [Which directory to use to chroot lldpd], [${runstatedir}/lldpd])
-lldp_ARG_WITH([lldpd-ctl-socket], [Path to socket for communication with lldpd], [${runstatedir}/lldpd.socket])
-lldp_ARG_WITH([lldpd-pid-file], [Path to lldpd PID file], [${runstatedir}/lldpd.pid])
+lldp_ARG_WITH([privsep-chroot], [Which directory to use to chroot lldpd], [${runstatedir}/lldpd/chroot])
+lldp_ARG_WITH([lldpd-ctl-socket], [Path to socket for communication with lldpd], [${runstatedir}/lldpd/lldpd.socket])
+lldp_ARG_WITH([lldpd-pid-file], [Path to lldpd PID file], [${runstatedir}/lldpd/lldpd.pid])
```
vincentbernat: This will be a pain for people upgrading. Why do you change them?

vincentbernat:
Hey!

I am OK with the change, except for the new locations. It may be cleaner to have everything in `/var/run/lldpd`, but this is not needed (the files outside are created while lldpd is running as root) and it will impact users upgrading. When using systemd, only the socket is in `/var/run`.

PR #341, opened 2019-09-03
Optional encapsulation format in Ethernet II and SNAP

rogerogerogeroger:
The encapsulation of LLDP protocal can be switched
between Ethernet II and SNAP.

Issue: 340
Change-Id: Ia010ddb72e959be9312f0de6eb2fe4a0586a8d0a


src/daemon/client.c
```diff
@@ -368,6 +368,13 @@ _client_handle_set_port(struct lldpd *cfg,
 		port->p_descr = strdup(set->local_descr);
 		port->p_descr_force = 1;
 	}
+	if (set->local_lldp_snap) {
+		log_debug("rpc", "Enable SNAP encapsulation format");
+		port->p_lldp_snap = 1;
+	} else if (set->local_lldp_snap == 0) {
+		log_debug("rpc", "Disable SNAP encapsulation format");
+		port->p_lldp_snap = 0;
+	}
```
vincentbernat: I think any port-related command will disable SNAP encapsulation. You should use 1 (true), 2 (false) instead.

src/lib/lldpctl.h
```diff
@@ -686,6 +686,7 @@ typedef enum {
 	lldpctl_k_port_id_subtype, /**< `(IS)` The subtype ID of this port.  */
 	lldpctl_k_port_id,	   /**< `(BS,WO)` The ID of this port. */
 	lldpctl_k_port_descr,	   /**< `(S,WO)` The description of this port. */
+	lldpctl_k_port_lldp_snap_enabled,	   /**< `(I,WO)` The lldp snap encapsulation enable status of this port. */
```
vincentbernat: Do not put a new value in the middle of existing values. It breaks the ABI (old clients cannot talk to new lldpd). Add it at the end of the section.

src/daemon/frame.h
```diff
@@ -64,6 +64,12 @@ static union {
 	pos = (where);				\
 	} while(0)
 
+#define POKE_REAL_LENGTH(value, type, bytes, func)		\
+	(type = func(value),                                    \
+	 memcpy(packet + 2*ETHER_ADDR_LEN, &type, bytes))
+#define POKE_REAL_LENGTH_REPLACE(value, bytes)			\
+        POKE_REAL_LENGTH(value, types.f_uint16, bytes, htons)
+
```
vincentbernat: These macros are used only in one place. I think this is better to put the code directly instead.

PR #367, opened 2019-11-21
LLDPD should document system refresh timer (tx-interval * 20)

stannous:
In LLDPD, each port has its own timer to catch port-related
changes and is modified by changing the tx-interval.

LLDPD also starts another system based refresh timer on each port
for changes like hostname. This is the tx-interval multiplied by
20.  This needs to be documented.

Signed-off-by: Sam Tannous <stannous@cumulusnetworks.com>


vincentbernat:
Thanks! I have done a minor edit.

PR #375, opened 2019-12-09
Fix for LLDP related netlink error messages

stannous:
This is a patch for https://github.com/vincentbernat/lldpd/issues/373

Issue: Error messages "netlink: 8 bytes leftover after parsing attributes
in process `lldpd'"

Root cause: Root Cause: The length of the netlink message was not being set
properly for non-bridge family type messages. Same length was being used for
both type of messages even though bridge family type message has extra
attribute. This causes 8 extra bytes being left over in the non-bridge
family type netlink messages.

Fix: Calculating and setting the length separately for bridge and non-bridge
family type messages.
(cherry picked from commit aac76966539bf932d5923b165762db370990bf94)

### DIFF
--- a/src/daemon/netlink.c
+++ b/src/daemon/netlink.c
@@ -145,7 +145,7 @@ netlink_send(int s, int type, int family, int seq)
 {
 	struct netlink_req req = {
 		.hdr = {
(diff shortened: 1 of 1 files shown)


vincentbernat:
Thanks, that's helpful!

PR #376, opened 2019-12-09
Fix for limitation of 10 VLANs for LLDP .1q feature

stannous:
Max 10 VLAN ids are supported on a port for .1q feature

Root Cause: All the VLANs learnt from netlink is dropped after 10 VLANs due
to the static array allocation of only 10 VLAN ids in the interface structure.
Beyond 10 VLAN membership for a port are ignored and error message gets
printed causing flooding of messages when hundreds of VLANs are configured.

Fix: Changed the static VLAN id array to VLAN id bitmap. With the bitmap all 4k
VLANs can be stored and learnt from netlink messages.
- Added a message to indicate when the LLDP packet is not sent out because
its too big. This will be helpful for user when too many VLANs are
configured and LLDP packets are not sent out.

Limitation: Even though the VLAN learning from netlink messages has been
alleviated, due to the LLDP message size around 380 VLANs can be advertised in
the packet. This number can vary based on the number of other TLVs being
advertised by LLDP.

vlan info shows interface_name.vlan-id which makes it look like sub
interface and causes confusion

Root Cause: Vlan name is sent as part of the .1q TLVs. But, the vlan name
format was <nterface-name>.<vlan-id> which makes it look like a sub-interface.

Fix: The vlan name cannot be removed from the vlan-info display since it is
sent/received as part of the .1q TLV in LLDP packet. But, changed the
vlan name format to vlan-<vlan-id> to avoid the confusion.
(cherry picked from commit 38db598121f5ce615f98d6cdaf41d5360c40dc3c)

Conflicts:
	src/daemon/interfaces-linux.c
	src/daemon/lldpd.h

### DIFF
--- a/tests/integration/test_interfaces.py
+++ b/tests/integration/test_interfaces.py
@@ -83,7 +83,7 @@ def test_vlan_aware_bridge_with_vlan(lldpd1, lldpd, lldpcli, namespaces, links,
         out = lldpcli("-f", "keyvalue", "show", "neighbors", "details")
         assert out['lldp.eth0.port.descr'] == 'eth1'
         assert out['lldp.eth0.vlan'] == \
-            ['eth1.100', 'eth1.200', 'eth1.300']
+            ['vlan100', 'vlan200', 'vlan300']
         assert out['lldp.eth0.vlan.vlan-id'] == \
             ['100', '200', '300']
         assert out['lldp.eth0.vlan.pvid'] == \
@@ -217,10 +217,7 @@ def test_remove_vlan(lldpd1, lldpd, lldpcli, namespaces, links, kind):
     with namespaces(1):
         out = lldpcli("-f", "keyvalue", "show", "neighbors", "details")
         assert out['lldp.eth0.port.descr'] == 'eth1'
-        if kind != 'vlan-aware-bridge':
-            assert out['lldp.eth0.vlan'] == ['vlan400', 'vlan500']
-        else:
-            assert out['lldp.eth0.vlan'] == ['eth1.400', 'eth1.500']
+        assert out['lldp.eth0.vlan'] == ['vlan400', 'vlan500']
         assert out['lldp.eth0.vlan.vlan-id'] == ['400', '500']
         assert out['lldp.eth0.vlan.pvid'] == ['no', 'no']
 
--- a/NEWS
+++ b/NEWS
@@ -7,6 +7,8 @@ lldpd (1.0.5)
     + Add support for VLAN-aware bridges for Linux (no range support).
     + Add support for 802.3BT (no SNMP support).
     + Add support for millisecond-grained tx-interval (Jean-Pierre Tosoni).
+    + Use generic names for VLAN names, instead of interface names (eg
+      vlan100 instead of eth1.100).
   * Fix:
     + Don't clear chassis TLV on shutdown LLDPDU.
     + Don't require/display powerpairs for Dot3 power when device type
--- a/src/daemon/interfaces-bsd.c
+++ b/src/daemon/interfaces-bsd.c
@@ -307,7 +307,7 @@ ifbsd_check_vlan(struct lldpd *cfg,
 	    "%s is VLAN %d of %s",
 	    vlan->name, vreq.vlr_tag, lower->name);
 	vlan->lower = lower;
-	vlan->vlanids[0] = vreq.vlr_tag;
+	bitmap_set(vlan->vlan_bmap, vreq.vlr_tag);
 	vlan->type |= IFACE_VLAN_T;
 }
 
--- a/src/daemon/interfaces-linux.c
+++ b/src/daemon/interfaces-linux.c
@@ -222,7 +222,7 @@ iflinux_is_vlan(struct lldpd *cfg,
 		}
 
 		iface->lower = lower;
-		iface->vlanids[0] = ifv.u.VID;
+		bitmap_set(iface->vlan_bmap, ifv.u.VID);
 		return 1;
 	}
 #endif
--- a/src/daemon/interfaces.c
+++ b/src/daemon/interfaces.c
@@ -212,6 +212,7 @@ iface_append_vlan(struct lldpd *cfg,
 	struct lldpd_port *port;
 	struct lldpd_vlan *v;
 	char *name = NULL;
+	uint16_t vlan_id;
 
 	if (hardware == NULL) {
 		log_debug("interfaces",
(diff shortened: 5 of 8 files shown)


vincentbernat:
The restriction to 10 VLAN was voluntary to avoid having a LLPDU too big, but the solution of using a bitmap is more elegant either way, so OK. Also OK for the name change.

Could you fix the issue reported in the tests. I think, you will also need to update some tests, but you can push the first fix and we'll see which tests are failing.

tests/integration/test_dot1.py
```diff
@@ -11,7 +11,7 @@ def test_one_vlan(self, lldpd1, lldpd, lldpcli, namespaces, links):
             lldpd()
         with namespaces(1):
             out = lldpcli("-f", "keyvalue", "show", "neighbors", "details")
-            assert out['lldp.eth0.vlan'] == 'vlan100'
+            assert out['lldp.eth0.vlan'] == 'vlan-100'
```
vincentbernat: This shouldn't change. The name of the interface is `vlan100`.

tests/integration/test_dot1.py
```diff
@@ -23,7 +23,7 @@ def test_several_vlans(self, lldpd1, lldpd, lldpcli, namespaces, links):
             out = lldpcli("-f", "keyvalue", "show", "neighbors", "details")
             # We know that lldpd is walking interfaces in index order
             assert out['lldp.eth0.vlan'] == \
-                ['vlan100', 'vlan200', 'vlan300', 'vlan4000']
+                ['vlan-100', 'vlan-200', 'vlan-300', 'vlan-4000']
```
vincentbernat: Same here.

tests/integration/test_interfaces.py
```diff
@@ -57,7 +57,7 @@ def test_bridge_with_vlan(lldpd1, lldpd, lldpcli, namespaces, links, when):
         out = lldpcli("-f", "keyvalue", "show", "neighbors", "details")
         assert out['lldp.eth0.port.descr'] == 'eth1'
         assert out['lldp.eth0.vlan'] == \
-            ['vlan100', 'vlan200', 'vlan300']
+            ['vlan-100', 'vlan-200', 'vlan-300']
```
vincentbernat: This is a classic bridge with VLAN on top of it. The name of interfaces are `vlan100`, ...

tests/integration/test_interfaces.py
```diff
@@ -83,7 +83,7 @@ def test_vlan_aware_bridge_with_vlan(lldpd1, lldpd, lldpcli, namespaces, links,
         out = lldpcli("-f", "keyvalue", "show", "neighbors", "details")
         assert out['lldp.eth0.port.descr'] == 'eth1'
         assert out['lldp.eth0.vlan'] == \
-            ['eth1.100', 'eth1.200', 'eth1.300']
+            ['vlan-100', 'vlan-200', 'vlan-300']
```
vincentbernat: This change is OK.

tests/integration/test_interfaces.py
```diff
@@ -160,7 +160,7 @@ def test_bond_with_vlan(lldpd1, lldpd, lldpcli, namespaces, links, when):
         out = lldpcli("-f", "keyvalue", "show", "neighbors", "details")
         assert out['lldp.eth0.port.descr'] == 'eth1'
         assert out['lldp.eth0.vlan'] == \
-            ['vlan300', 'vlan301', 'vlan302', 'vlan303']
+            ['vlan-300', 'vlan-301', 'vlan-302', 'vlan-303']
```
vincentbernat: This shouldn't change either.

tests/integration/test_interfaces.py
```diff
@@ -181,7 +181,7 @@ def test_just_vlan(lldpd1, lldpd, lldpcli, namespaces, links, when):
     with namespaces(1):
         out = lldpcli("-f", "keyvalue", "show", "neighbors", "details")
         assert out['lldp.eth0.port.descr'] == 'eth1'
-        assert out['lldp.eth0.vlan'] == ['vlan300', 'vlan400']
+        assert out['lldp.eth0.vlan'] == ['vlan-300', 'vlan-400']
```
vincentbernat: Should not change.

tests/integration/test_interfaces.py
```diff
@@ -216,9 +216,9 @@ def test_remove_vlan(lldpd1, lldpd, lldpcli, namespaces, links, kind):
         out = lldpcli("-f", "keyvalue", "show", "neighbors", "details")
         assert out['lldp.eth0.port.descr'] == 'eth1'
         if kind != 'vlan-aware-bridge':
-            assert out['lldp.eth0.vlan'] == ['vlan400', 'vlan500']
+            assert out['lldp.eth0.vlan'] == ['vlan-400', 'vlan-500']
         else:
-            assert out['lldp.eth0.vlan'] == ['eth1.400', 'eth1.500']
+            assert out['lldp.eth0.vlan'] == ['vlan-400', 'vlan-500']
```
vincentbernat: Ok for the `if` branch, but not for the `else` branch.

tests/integration/test_interfaces.py
```diff
@@ -252,7 +252,7 @@ def test_unenslave_bond_with_vlan(lldpd1, lldpd, lldpcli, namespaces, links):
     with namespaces(1):
         out = lldpcli("-f", "keyvalue", "show", "neighbors", "details")
         assert out['lldp.eth0.port.descr'] == 'eth1'
-        assert out['lldp.eth0.vlan'] == 'vlan400'
+        assert out['lldp.eth0.vlan'] == 'vlan-400'
```
vincentbernat: Should not change.

tests/integration/test_interfaces.py
```diff
@@ -284,7 +284,7 @@ def test_down_then_up_with_vlan(lldpd1, lldpd, lldpcli, namespaces, links):
     with namespaces(1):
         out = lldpcli("-f", "keyvalue", "show", "neighbors", "details")
         assert out['lldp.eth0.port.descr'] == 'eth1'
-        assert out['lldp.eth0.vlan'] == ['vlan300', 'vlan400']
+        assert out['lldp.eth0.vlan'] == ['vlan-300', 'vlan-400']
```
vincentbernat: Should not change.

vincentbernat:
Maybe it would be simpler to split this patch in two:
1. Remove the 10 limitation
2. Name VLAN for a VLAN-aware bridge differently

vincentbernat:
After thinking a bit around this, I have decided it may be better to always name VLANs "vlanX" instead of using sometimes the interface name (which could happen to be `vlan100` for old-style VLANs) and sometimes building from interface name and VLAN ID. So, it is similar to what you did, except I have removed the hyphen.

I have amended and squashed your commits. I have pushed some extra commits:
 - fix detection of VLAN 1 (VLAN 1 encoded as `2` when alone)
 - use a well-known optimized version for `num_bits_set` (though it could be optimized even more as we only care VLAN 1, but let's keep it generic for now)

Tell me if you are OK with this.

PR #377, opened 2019-12-09
Fix for extra VLANs from bridge sent in the LLDP message

stannous:
Extra VLANs that the bridge port is not part of are being advertised in LLDP message to the peer

Root Cause: The bridge member port lldp messages also includes all the VLANs
            that its master bridge in part of even though the port is not part
            of those VLANs. This happens because during netlink VLAN updates
            all the bridge members of a bridge are traversed and VLANs on the
            bridge are applied to its bridge members.

Fix: Added a check to ignore the bridge interfaces during processing/applying
     of VLAN updates from netlink messages and not propagate the VLANs
     configured against the bridges to its members.
(cherry picked from commit 607015aa407c4ef7c601d8cdbb801ab9e8a53eb1)

### DIFF
--- a/src/daemon/interfaces.c
+++ b/src/daemon/interfaces.c
@@ -285,6 +285,14 @@ iface_append_vlan_to_lower(struct lldpd *cfg,
 	    "looking to apply VLAN %s to physical interface behind %s",
 	    vlan->name, upper->name);
 
+	/* Ignore the bridge interface since the bridge member interfaces may not
+	 * inherit all VLANs the bridge interface has */
+	if (upper->type & IFACE_BRIDGE_T) {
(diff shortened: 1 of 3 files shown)


vincentbernat:
But then, it breaks the case where people add VLAN on top of a "classic" bridge. This is tested in `test_bridge_with_vlan()`.

vincentbernat:
Also, please give me the output of `bridge vlan` for which you try to fix the behaviour of lldpd. This way, we can add a test.

vincentbernat:
I think you have appended your commit to the wrong pull request?

Moreover, you should not change the tests around non-VLAN-aware bridges. If you need it, it means there is a problem in the code.

vincentbernat:
Let's focus on #376 first.

PR #379, opened 2020-01-16
Allow configuring tx-interval in milliseconds

jp-t:
This is a prerequisite for TTDP support.

Usage: add "ms" suffix to the delay value to force interpretation as a
milliseconds value:

 lldpcli configure lldp tx-interval 200ms

When computing a TTL (TTL = tx-interval * tx-hold), the result is rounded
up to the next second, as LLDP frames need a value in seconds.

A new tx-interval-ms item is added to the json output. The old tx-interval
is kept in seconds (rounded up) for compatibility.

### DIFF
--- a/src/daemon/event.c
+++ b/src/daemon/event.c
@@ -483,6 +483,8 @@ levent_update_and_send(evutil_socket_t fd, short what, void *arg)
 	lldpd_loop(cfg);
 	if (cfg->g_iface_event != NULL)
 		interval_ms *= 20;
+	if (interval_ms < 30000)
+		interval_ms = 30000;
 	tv.tv_sec = interval_ms / 1000;
 	tv.tv_usec = (interval_ms % 1000) * 1000;
 	event_add(cfg->g_main_loop, &tv);
--- a/src/daemon/event.c
+++ b/src/daemon/event.c
@@ -476,11 +476,15 @@ static void
 levent_update_and_send(evutil_socket_t fd, short what, void *arg)
 {
 	struct lldpd *cfg = arg;
-	struct timeval tv = { cfg->g_config.c_tx_interval, 0 };
+	struct timeval tv;
+	long interval_ms = cfg->g_config.c_tx_interval;
+
 	(void)fd; (void)what;
 	lldpd_loop(cfg);
 	if (cfg->g_iface_event != NULL)
-		tv.tv_sec *= 20;
+		interval_ms *= 20;
+	tv.tv_sec = interval_ms / 1000;
+	tv.tv_usec = (interval_ms % 1000) * 1000;
 	event_add(cfg->g_main_loop, &tv);
 }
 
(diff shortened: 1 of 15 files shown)


src/daemon/event.c
```diff
@@ -476,11 +476,15 @@ static void
...
+	struct timeval tv;
+	long interval_ms = cfg->g_config.c_tx_interval;
+
 	(void)fd; (void)what;
 	lldpd_loop(cfg);
 	if (cfg->g_iface_event != NULL)
-		tv.tv_sec *= 20;
+		interval_ms *= 20;
+	tv.tv_sec = interval_ms / 1000;
+	tv.tv_usec = (interval_ms % 1000) * 1000;
```
vincentbernat: This is a safeguard to refresh information about all network interfaces at regular interval, in case there was something wrong with the event-based refresh. Maybe you can add a minimal value of 1 minute here.

vincentbernat: Minimal value of 30 seconds even if we don't have event-based refresh (so outside the `if`). Most people have the event-based refresh, so lowering the pace for the others should be pretty invisible.

jp-t:
> 
> 
> At first, I think this is good. Some small points:
> 
>     * Could you add a test in `test_lldpcli.py` as well (just like existing `tx-interval` test)?
>     * Could you update the manual page for lldpcli?
>     * Could you add an entry in NEWS (credit yourself if you want to)?

I have done all this, but when running pytest, I have an error that I think is unrelated with my work:
```
[gw5] [ 98%] FAILED test_snmp.py::test_snmp_register

and later

>           assert 'STRING: "lldpMIB implementation by lldpd"' in out.values()
E           assert 'STRING: "lldpMIB implementation by lldpd"' in dict_values(['STRING: "View-based Access Control Model for SNMP."'])
E            +  where dict_values(['STRING: "View-based Access Control Model for SNMP."']) = <built-in method values of dict object at 0x7f5e7d279b40>()
E            +    where <built-in method values of dict object at 0x7f5e7d279b40> = {'.1.3.6.1.2.1.1.9.1.3.1': 'STRING: "View-based Access Control Model for SNMP."'}.values

test_snmp.py:13: AssertionError

````
Any hint ? I am not in Python at all... so it's hard to debug further

**Comment by vincentbernat** [comment]
Try to push your modifications here so that Travis run the tests. I get the same error from time to time but I think Travis is always OK. If that's the only error, it should be OK.

File: src/daemon/event.c (hunk 1 above)
Comment (jp-t) [reply src/daemon/event.c]: Do you mean something like this ?

```
@@ -483,6 +483,8 @@ levent_update_and_send(evutil_socket_t fd, short what, void *arg)
        lldpd_loop(cfg);
        if (cfg->g_iface_event != NULL)
                interval_ms *= 20;
+       if (interval_ms < 30000)
+               interval_ms = 30000;
        tv.tv_sec = interval_ms / 1000;
        tv.tv_usec = (interval_ms % 1000) * 1000;
        event_add(cfg->g_main_loop, &tv);

```

vincentbernat: Yes!

jp-t: I just pushed this change, I quoted you in the commit message, are you OK with this ?

vincentbernat:
Yes, it's OK. I am restarting the failed test. This seems unrelated to
your commit.

jp-t:
I guess I should also update the man page saying that "...tx-interval multiplied by 20..." to say that there is a lower boundary of 30s now ?

vincentbernat:
I don't think it is worth adding more documentation for that.

PR #404, opened 2020-07-05
lib: fix enum value

awujek:
Signed-off-by: Adam Wujek <dev_public@wujek.eu>

### DIFF
--- a/src/lib/lldpctl.h
+++ b/src/lib/lldpctl.h
@@ -842,7 +842,7 @@ typedef enum {
 	lldpctl_k_med_power_val, /**< `(I,W)` LLDP MED power value */
 
 	lldpctl_k_mgmt_ip = 3000,	/**< `(S)` IP address */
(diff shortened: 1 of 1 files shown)


src/lib/lldpctl.h
```diff
@@ -842,7 +842,7 @@ typedef enum {
 	lldpctl_k_med_power_val, /**< `(I,W)` LLDP MED power value */
 
 	lldpctl_k_mgmt_ip = 3000,	/**< `(S)` IP address */
-	lldpctl_k_mgmt_iface_index = 30001,	/**< `(I)` Interface index */
```
awujek, line 845: IMHO: there is one zero too much in "= 30001". If so, it is not needed.

vincentbernat:
I don't know what I did this. Unfortunately, it would be an ABI break to change this value. I'll try to remember next time there is a major change that would trigger an ABI break.

Issue #403, opened 2020-07-03
send LLDP frames inside VLAN


awujek:
Hi all, 

I would like to ask if it is possible to send LLDP frames tagged with VLAN. I managed to achieve the expected behaviour by creating a VLAN interface in linux by:
`ip link add link eth0 name eth0.14 type vlan id 14`
then running lldpd:
`lldpd -I eth0.14`

However, I prefer not to create interface for VLAN. I expect that lldpd adds a VLAN tag directly at the send of a frame.
I would like to see an option like:
`configure ports eth0 lldp vlan 14 prio 1`

I have understood that the configuration of med does not cause the tagging of lldp frames. Is that correct?
I see that lldpd processes lldp frames no matter if they have VLAN tag or not.

vincentbernat:
Processing LLDP frames inside VLAN is done to workaround historical bugs where some equipments (old Cisco Catalyst) were encapsulating LLDP frames into the native VLAN.

Technically, your request is possible and not too difficult if you only specify one VLAN per port. I don't have myself much time to implement that.

awujek:
Implemented

PR #410, opened 2020-07-20
#403 ports: make possible sending LLDP frames on a specified VLAN

awujek:
The current limitation is that frames are sent only on one VLAN per port.

### DIFF
--- a/src/lib/lldpctl.h
+++ b/src/lib/lldpctl.h
@@ -738,6 +738,7 @@ typedef enum {
 	lldpctl_k_port_status,	   /**< `(IS,WO)` Operational status of this (local) port */
 	lldpctl_k_port_chassis,	   /**< `(A)` Chassis associated to the port */
 	lldpctl_k_port_ttl,        /**< `(I)` TTL for port, 0 if info is attached to chassis */
+	lldpctl_k_port_vlan_tx,    /**< `(I,W)` VLAN tag for TX on port, -1 VLAN disabled */
 
 	lldpctl_k_port_dot3_mfs = 1300,	   /**< `(I)` MFS */
 	lldpctl_k_port_dot3_aggregid,   /**< `(I)` Port aggregation ID */
--- a/NEWS
+++ b/NEWS
@@ -8,6 +8,7 @@ lldpd (1.0.6)
     + Deprecate use of lldpctl_watch_callback(). Use
       lldpctl_watch_callback2() instead.
     + Upgrade embedded libevent to 2.1.11-stable
+    + Add support of sending LLDP frames on a configured VLAN
 
 lldpd (1.0.5)
   * Changes:
--- a/src/client/conf-lldp.c
+++ b/src/client/conf-lldp.c
@@ -313,6 +313,85 @@ cmd_chassis_mgmt_advertise(struct lldpctl_conn_t *conn, struct writer *w,
 	return 1;
 }
 
+static int
+cmd_vlan_tx(struct lldpctl_conn_t *conn, struct writer *w,
+		struct cmd_env *env, void *arg)
+{
+	lldpctl_atom_t *port;
+	const char *name;
+	const char *vlan_id = cmdenv_get(env, "vlan-tx-id");
+	const char *vlan_prio = cmdenv_get(env, "vlan-tx-prio");
+	const char *vlan_dei = cmdenv_get(env, "vlan-tx-dei");
+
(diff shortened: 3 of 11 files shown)


src/daemon/lldpd.c
```diff
@@ -170,6 +170,9 @@ lldpd_alloc_default_local_port(struct lldpd *cfg)
 		calloc(1, sizeof(struct lldpd_port))) == NULL)
 		fatal("main", NULL);
 
+	/* Initialize TX VLAN */
```
awujek, line 173: Is this the correct place to initialize port->p_vlan_tx_id?

src/lib/lldpctl.h
```diff
@@ -738,6 +738,9 @@ typedef enum {
 	lldpctl_k_port_status,	   /**< `(IS,WO)` Operational status of this (local) port */
 	lldpctl_k_port_chassis,	   /**< `(A)` Chassis associated to the port */
 	lldpctl_k_port_ttl,        /**< `(I)` TTL for port, 0 if info is attached to chassis */
+	lldpctl_k_port_vlan_tx_id, /**< `(I)` VLAN ID for TX on port, -1 VLAN disabled */
+	lldpctl_k_port_vlan_tx_prio, /**< `(I)` Priority of VLAN for TX on port, default 0 */
+	lldpctl_k_port_vlan_tx_dei, /**< `(I)` Drop eligible indicator (DEI) of VLAN for TX on port, default 0 */
```
vincentbernat, line 743: They can also be written, so they should be marked as `(I,W)`.

src/daemon/lldpd.c
```diff
@@ -201,6 +204,10 @@ lldpd_clone_port(struct lldpd_port *destination, struct lldpd_port *source)
 	memcpy(destination, cloned, sizeof(struct lldpd_port));
 	free(cloned);
 	free(output);
+
+	/* Initialize TX VLAN */
+	destination->p_vlan_tx_id = -1;
```
vincentbernat, line 209: I don't think you should initialize here. This would prevent applying a VLAN to all ports. The default port is a template for all future ports, to ensure they get a configuration applied to all ports.

src/daemon/lldpd.c
vincentbernat: Yes, it looks good.

vincentbernat:
The code looks good. Some unittests are failing, I didn't have a look yet.

Tell me if you plan to write the manual page and add integration tests (I can do the later part if needed).

src/daemon/lldpd.c
awujek: Could you give ma a hint where shall `p_vlan_tx_id` be initialized?

awujek:
Hi Vincent,

Thanks a lot for reviewing my changes!
> The code looks good. Some unittests are failing, I didn't have a look yet.

Some unittests are failing because `p_vlan_tx_id` is initialized in a wrong place. I will fix tests as well.
> 
> Tell me if you plan to write the manual page and add integration tests (I can do the later part if needed).

Yes, I will write the manual page and integration tests, I just wanted your feedback whether the general idea is correct.

src/lib/lldpctl.h
awujek: good point, I forgot about this.

src/daemon/lldpd.c
vincentbernat: The other place looks good (`lldpd_alloc_default_local_port()`)

awujek:
TODO: manual+tests

src/daemon/lldpd.c
awujek: There is no need to have initialization any more. I introduced `p_vlan_tx_enabled`, which is `1` when the tag is valid.

src/lib/lldpctl.h
awujek: I left only one `lldpctl_k_port_vlan_tx`. Which has `(I,W)` now.
Done.

vincentbernat:
This looks good. Let me add a functional test sending an actual frame. I'll do that shortly.

awujek:
> This looks good. Let me add a functional test sending an actual frame. I'll do that shortly.

There is a test already `test_send_rcv_vlan_tx`, which assembles the frame then strips the VLAN tag (since it can be done by Linux kernel) and decodes it.

vincentbernat:
OK then, let's merge!

PR #418, opened 2020-11-17
lldp: avoid memory leak from bad packets

orgcandman:
A packet that contains multiple instances of certain TLVs will cause
lldpd to continually allocate memory and leak the old memory.  As an
example, multiple instances of system name TLV will cause old values
to be dropped by the decoding routine.

Reported-at: https://github.com/openvswitch/ovs/pull/337
Reported-by: Jonas Rudloff <jonas.t.rudloff@gmail.com>
Signed-off-by: Aaron Conole <aconole@redhat.com>


vincentbernat:
Looks good! Thanks!

PR #419, opened 2020-12-03
lldpd: mgmt-ip info showing invalid ip in lldp neighbors output

sudhanshukumar22:
Problem Description/Summary :
--------------------------------------------
mgmt-ip info showing invalid ip in lldp neighborship output.

Expected Behavior :
------------------------
mgmt-ip should be same in both show lldp neighbors and ifconfig eth0

Topology :
------------
2DUT

Fix: Update the management ip.
Verified LLDP neighbor information has the mgmt interface IP only, verified reboot and docker restart cases as well.
Signed-off-by: sudhanshukumar22 <sudhanshu.kumar@broadcom.com>

### DIFF
--- a/src/daemon/netlink.c
+++ b/src/daemon/netlink.c
@@ -26,6 +26,7 @@
 #include <linux/netlink.h>
 #include <linux/rtnetlink.h>
 #include <linux/if_bridge.h>
+#include <net/if.h>
 
 #define NETLINK_BUFFER 4096
 
@@ -424,10 +425,30 @@ netlink_parse_address(struct nlmsghdr *msg,
 	struct ifaddrmsg *ifi;
 	struct rtattr *attribute;
 	int len;
+	char ifname[IFNAMSIZ];  
+	static int mgmt_ifaindex = -1;
 	ifi = NLMSG_DATA(msg);
 	len = msg->nlmsg_len - NLMSG_LENGTH(sizeof(struct ifaddrmsg));
 
 	ifa->index = ifi->ifa_index;
+	if(mgmt_ifaindex == -1)     
+	{                               
+		if_indextoname(ifi->ifa_index, ifname);
+		if(strncmp(ifname, "eth0", 4) == 0)
+		{       
+			mgmt_ifaindex = ifa->index;
+		}
+	}
+
+	if(mgmt_ifaindex != ifa->index)
+	{
+		log_debug("netlink", "non mgmt IP address on if %d (family: %d)",
+				ifa->index, ifi->ifa_family);
+		return 0;
+	}
+
+	log_info("netlink", "mgmt IP address on if %d (family: %d)",
+		ifa->index, ifi->ifa_family);
 	ifa->flags = ifi->ifa_flags;
 	switch (ifi->ifa_family) {
 	case AF_INET:


vincentbernat:
You can configure the management IP using `configure system ip management pattern`. Depending on the system, this could be the one on `eth0` or not. Hard-coding `eth0` is not possible.

vincentbernat:
lldpd runs on any Linux/BSD. One some of them, the management interface will be on `vlan100`, on others `em0`, on others `eno3`. There is no universal rule.

I understand the management pattern is not enough in your case. I can introduce a way to specify an interface name instead of an IP address. In this case, you can ship lldpd with the option set to `eth0`. Would this matches your use case?

sudhanshukumar22:
> lldpd runs on any Linux/BSD. One some of them, the management interface will be on `vlan100`, on others `em0`, on others `eno3`. There is no universal rule.
> 
> I understand the management pattern is not enough in your case. I can introduce a way to specify an interface name instead of an IP address. In this case, you can ship lldpd with the option set to `eth0`. Would this matches your use case?

Do you want to introduce another CLI  to specify the interface name for management interface ? I think that should help, but problem is user may not enter that CLI. In that case, we have to take the ip address from some fixed interface (like eth0).

vincentbernat:
❦  3 décembre 2020 05:44 -08, sudhanshukumar22:

>> I understand the management pattern is not enough in your case. I
>> can introduce a way to specify an interface name instead of an IP
>> address. In this case, you can ship lldpd with the option set to
>> `eth0`. Would this matches your use case?
>
> Do you want to introduce another CLI to specify the interface name for
> management interface ?

Yes.

> I think that should help, but problem is user may not enter that CLI.

If you distribute lldpd as part of your platform, you can ship a
/etc/lldpd.d/lldpd.conf with the command preinstalled. The user can
override it if they want.

> In that case, we have to take the ip address from some fixed interface
> (like eth0).

There are so many systems without an eth0 interface that it does not
seem the right heuristic. The current heuristic is to use the first IP
address. This heuristic is more than 10 years old.

I think it should work in your case too if eth0 is the first interface
on your system. However, if you use DHCP for eth0 and have static IP,
lldpd will not change back to the IP on eth0 once it gets a management
interface. A work-around would be to restart lldpd on DHCP leases.

sudhanshukumar22:
@vincentbernat : "However, if you use DHCP for eth0 and have static IP,
lldpd will not change back to the IP on eth0 once it gets a management
interface. A work-around would be to restart lldpd on DHCP leases."

By static IP, do you mean that user has configured ip address in LLDP using "configure system ip management pattern" ? Also, I don't understand "lldpd will not change back to the IP". is it because of CLI configuration in LLDP ?

vincentbernat:
❦  3 décembre 2020 06:11 -08, sudhanshukumar22:

> @vincentbernat : "However, if you use DHCP for eth0 and have static IP,
> lldpd will not change back to the IP on eth0 once it gets a management
> interface. A work-around would be to restart lldpd on DHCP leases."
>
> By static IP, do you mean that user has configured ip address in LLDP
> using "configure system ip management pattern" ? Also, I don't
> understand "lldpd will not change back to the IP". is it because of
> CLI configuration in LLDP ?

For example, if eth0 is using DHCP, but you have swp1 with a static IP
address (set outside lldpd), lldpd will select swp1 IP address as the
management IP address as this is the first one. When eth0 gets its DHCP
address, lldpd will not switch back to the IP address of eth0 because it
won't change the management IP address once it has one. However, once
eth0 has its IP address, if you restart lldpd, it will likely use the
one from eth0 (it really depends if it's the first in the output of "ip
a s").

sudhanshukumar22:
@vincentbernat : I assume that you want this fix to be reworked as follows:
1. Introduce a lldp cli to set the management interface, where we configure the ip address. 
2. add a line in /etc/lldpd.d/lldpd.conf

vincentbernat:
For 1, the CLI would be to specify the interface to use to take the management address. There is already a CLI to force the IP address. I can do that if you are not in a hurry.

sudhanshukumar22:
My point was this PR has to be discarded. Is that correct ?

vincentbernat:
We can keep it like an issue until we fix it. But, yes, the code part of this PR can be discarded.

vincentbernat:
Tell me if you plan to implement it, so that both of us don't work on it at the same time.

Thanks.

sudhanshukumar22:
Hi, We will take it up.

sudhanshukumar22:
> Tell me if you plan to implement it, so that both of us don't work on it at the same time.
> 
> Thanks.



> Tell me if you plan to implement it, so that both of us don't work on it at the same time.
> 
> Thanks.

@vincentbernat : Hi Vincent, after having a discussion in the team and considering other priorities, can you please take up this issue ?

vincentbernat:
❦ 22 décembre 2020 05:42 -08, sudhanshukumar22:

>> Tell me if you plan to implement it, so that both of us don't work on it at the same time.
>> 
>> Thanks.
>
> @vincentbernat : Hi Vincent, after having a discussion in the team and
> considering other priorities, can you please take up this issue ?

Yes, I can, allow me a couple of weeks.

vincentbernat:
So, while trying to implement it, I have discovered this is already supported. You can either run `lldpd` with `-m eth0` or you can use `configure system ip management pattern eth0`.

sudhanshukumar22:
@vincentbernat : I hope the work for this is more than implementing the CLI. The work may involve changing the lldpd.conf file to configure the default interface as eth0 (we can use the same CLI mentioned in your comments).

vincentbernat:
As said previously, I won't change the default to `eth0` as it would break existing setups and many devices don't use `eth0`, including Linux devices.

PR #420, opened 2020-12-03
lldpd:Device hangs after running LLDP suite in Defensics fuzz tool

sudhanshukumar22:
Running Defensics LLDP suite often hangs the device and only way to retrieve it back is by power cycling the device.

As the device hangs and there is no access to Console/Mgmt Sessions, not able to take tech-support output in problem state.

Attaching the pcap files of LLDP packets from Defensics server, as the  report doesn't contain any failures.

Also attaching the syslog output before/at the time hitting problem state.

Additionally, I checked if there are any sporadic CPU/Memory hike while running and did not notice any.

Only once seen lldp crash and attaching the same as well.
Fix: added a validation check in MED processing in lldp_decode().
Signed-off-by: sudhanshukumar22 <sudhanshu.kumar@broadcom.com>

### DIFF
--- a/src/daemon/protocols/lldp.c
+++ b/src/daemon/protocols/lldp.c
@@ -637,8 +637,10 @@ lldp_decode(struct lldpd *cfg, char *frame, int s,
 	const char dcbx[] = LLDP_TLV_ORG_DCBX;
 	unsigned char orgid[3];
 	int length, gotend = 0, ttl_received = 0;
+	int lci_len;
 	int tlv_size, tlv_type, tlv_subtype, tlv_count = 0;
 	u_int8_t *pos, *tlv;
+	u_int8_t *med_loc_address;
 	char *b;
 #ifdef ENABLE_DOT1
 	struct lldpd_vlan *vlan = NULL;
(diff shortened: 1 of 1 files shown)


vincentbernat:
Thanks for your patch. However, it only works when the location is a civic address. The two other possibilities (coordinate and ELIN) does use the first byte as a length. In `lldp.c`, no decoding is done, so an incorrect value shouldn't matter. Also, looking in `atom/med.c`, I see the length is just ignored (we rely on the len of the TLV, which is not totally correct). Do you have a PCAP showing a packet exhibiting a problem?

sudhanshukumar22:
> Thanks for your patch. However, it only works when the location is a civic address. The two other possibilities (coordinate and ELIN) does use the first byte as a length. In `lldp.c`, no decoding is done, so an incorrect value shouldn't matter. Also, looking in `atom/med.c`, I see the length is just ignored (we rely on the len of the TLV, which is not totally correct). Do you have a PCAP showing a packet exhibiting a problem?

I am attaching the lldp packet capture:
[mgmt ip change issue.zip](https://github.com/lldpd/lldpd/files/5636441/mgmt.ip.change.issue.zip)

vincentbernat:
Thanks, I am running into a bit of trouble to run some tests. Hopefully, this will be solved next week.

vincentbernat:
If you can run again the test suite, it will be helpful to catch remaining bugs!

sudhanshukumar22:
@vincentbernat : Does your pull request 422 fix this issue ?

vincentbernat:
Yes, it should. I was able to reproduce your issue.

PR #423, opened 2020-12-09
build: prevent conflict with official AX_LIB_READLINE macro

jow-:
On systems where the official AX_LIB_READLINE (ax_lib_readline.m4) is
present in a globally shared autoconf include directory, auto(re)conf
will prefer including that offical version over the local variant due
to the offical macro having a higher serial number.

As a consequence, @READLINE_LIBS@ will not be substituted in *.in files,
eventually failing the compilation with errors similar to:

    gcc: error: READLINE_LIBS@: No such file or directory

Avoid this problem by renaming the incompatible local macro to
AX_LIB_READLINE_LLDPD which is sufficient to prevent any clashes.

We encountered this problem on OpenWrt which uses GNU autoconf-archive
to provide commonly used M4 macros through a global include directory,
which happens to ship AX_LIB_READLINE as well.

Signed-off-by: Jo-Philipp Wich <jo@mein.io>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -267,7 +267,7 @@ AC_ARG_WITH([readline],
   [],
   [with_readline=auto])
 if test x"$with_readline" != x"no"; then
-   AX_LIB_READLINE
+   AX_LIB_READLINE_LLDPD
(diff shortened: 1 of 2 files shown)


vincentbernat:
Thanks!

PR #430, opened 2021-01-20
lldpd: override compiler and linker option strings

mobbitster:
Compiler and linker option could contain absolute path, so
replace them with disclaimer. Those strings used only as
debug information to show not actual compilation options.

Signed-off-by: Oleksiy Obitotskyy <oobitots@cisco.com>


vincentbernat:
Thanks!

PR #434, opened 2021-03-08
Fix relative include paths to be more correct and portable.

jlindgren90:
Some relative include paths in subdirectories (src/daemon/protocols
and src/lib/atoms) were written relative to the parent directories
(src/daemon and src/lib).  This was okay in automake builds but
caused errors when porting to other build systems (for example,
Android make).


vincentbernat:
It looks good, thanks!

Issue #438, opened 2021-03-25
Issue on MacOS postinstall script for _lldpd user creation


AnthoBalitrand:
# Bug description

There's an issue on MacOS lldpd postinstall script, making install impossible on MacOS Big Sur. 
The logic used to find the next available UID/GID does not take into account an existing group without user on it. 

## Steps to reproduce the problem

 1. Run the postinstall script manually on MacOS Big Sur

 2. The following error is displayed : 

```
<main> attribute status: eDSRecordAlreadyExists
<dscl_cmd> DS Error: -14135 (eDSRecordAlreadyExists)
```

 3. In this case, this is because de UID/GID found by the postinstall script for _lldpd is 201. But in fact there's an existing group using GID 201, with no user on it : 

```
dscl . list /Users gid | grep 201
dscl . list /Groups gid | grep 201 
_guest                           201
```
 
## Expected outcome

As the postinstall script tries to find a UID/GID which are identical, it should also use the 
```
dscl . list /Groups gid command
```
command to find existing groups without users, and not use only the Users list.

vincentbernat:
Thanks for the report! I have updated to use "

PR #439, opened 2021-03-25
Compare UID with existing GID without users

AnthoBalitrand:
Solves issue #438


vincentbernat:
Sorry, I didn't see your PR in time!

PR #350, opened 2019-09-17
lldp: Power via MDI TLV: Initial support for 802.3bt

patrikdanielssonaxis:
Initial support for sending and receiving the new frames specified in 802.3bt. The type_ext field is used to differentiate 802.3bt, with zero value meaning no bt support.

Tested and verified in conjunction with a Cisco Catalyst 9300 switch.

PR #449, opened 2021-05-18
src/client/conf-power.c: Add support for 803.2bt parameters.

andreas-hofmann:
This patch adds support for the 802.3bt parameters to lldpcli.

### DIFF
--- a/src/client/conf-power.c
+++ b/src/client/conf-power.c
@@ -22,6 +22,13 @@
 #include "../log.h"
 
 static int
+cmd_check_ext_type(struct cmd_env *env, const char *type);
+static int
+cmd_check_ext_pse_but_no(struct cmd_env *env, void *arg);
+static int
+cmd_check_ext_pd_but_no(struct cmd_env *env, void *arg);
+
+static int
 cmd_medpower(struct lldpctl_conn_t *conn, struct writer *w,
     struct cmd_env *env, void *arg)
 {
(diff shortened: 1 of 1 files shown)


vincentbernat:
Thanks! This looks good. Could you add a test in `tests/integration/test_dot3.py`?

bryceremick:
@andreas-hofmann Hello, what is the status of this PR? I've tested it and an issue I'm seeing is that
`requested-a`, `requested-b`, `allocated-a`, `allocated-b`, `pd-load` are not optional, which is contrary to what was suggested in this [comment](https://github.com/lldpd/lldpd/pull/350#issuecomment-546923233) (which is what you based this PR off of)

d1115choi:
Hi all,
I am wondering why the PR is not merged yet or what things are not resolved?

vincentbernat:
I was waiting for tests to be added. I suppose I could merge as is and add tests myself.

andreas-hofmann:
Hi @vincentbernat (and everyone else for that matter). I'm sorry for pretty much abandoning this PR - I developed this on my old job, but after some requirements changed it was not needed and due to private reasons I obviously did not get around to actually finish it off with some tests. I hope this still helped some people out there, and if you'd like to merge it, I'd of course be happy, too.
Sorry again for any inconvenience caused.

vincentbernat:
I have rebased the code against master. I'll look into adding tests.

vincentbernat:
After trying to add some tests, I discovered that 802.3bt support in lldpd was broken since the beginning. I don't know exactly why, but the TLV generated had an incorrect length. After digging a bit more, it seems Wireshark does not agree on everything present in the TLV. Maybe the support introduced in #350 was good enough to parse, but the sending part was not tested? Or maybe it's based on an earlier version of the standard? As the standard is not freely available, it is difficult for me to check. I am not comfortable maintaining something broken. The work previously done is still present in the history in case someone want to dig a bit.

vincentbernat:
I've added back the parsing part of 802.3bt.

ShroudedNight:
In re spec availability: The 802.3bt TLV definitions appear to be included in the complete 802.3-2022 spec roll-up available gratis as part of the [IEEE "GET Program"](https://ieeexplore.ieee.org/browse/standards/get-program/page) under section 79 "IEEE 802.3 Organizationally Specific Link Layer Discovery Protocol (LLDP) type, length, and value (TLV) information elements"

Ekmansoft:
> After trying to add some tests, I discovered that 802.3bt support in lldpd was broken since the beginning.

Can you elaborate on what problems you were seeing? 

I am testing it now towards a Sifos PDA604 and toward a Netgear MS108TUP and I think the data from the PD looks ok when decoded in Wireshark when running version 1.0.17 (which should be before the bt support was removed, as I understand it).

PR #456, opened 2021-07-21
Fix hardcoded ghcr.io path, add missing docker archs

berlin4apk:
Fix hardcoded ghcr.io path
ghcr.io/${{ github.repository }}

add missing docker archs
            linux/386
            linux/amd64
            linux/arm64
            linux/arm/v6
            linux/arm/v7
            linux/arm64
            linux/ppc64le
            linux/s390x

And fix the push at a PR

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,6 +112,6 @@ jobs:
             linux/amd64
             linux/arm64
             linux/arm/v6
-          push: true
+          push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,8 @@ jobs:
       - uses: actions/checkout@v2
       - uses: docker/setup-qemu-action@v1
       - uses: docker/setup-buildx-action@v1
+      - name: Available platforms
+        run: echo ${{ steps.buildx.outputs.platforms }}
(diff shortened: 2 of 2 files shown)


berlin4apk:
Feedback are welcome, if it's looks so good.

.github/workflows/ci.yml
```diff
@@ -98,9 +109,13 @@ jobs:
...
+            linux/386
             linux/amd64
-            linux/arm64
             linux/arm/v6
-          push: true
+            linux/arm/v7
+            linux/arm64
+            linux/ppc64le
+            linux/s390x
+          push: ${{ github.event_name != 'pull_request' }}
```
vincentbernat, line 119: There is a check above: `github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/')`, so it can't match on pull request I think. If you modify it, be sure to not allow Docker push on push event, as I usually create my own branches in the same repository.

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
```
vincentbernat, line 9: Is it for the generated Docker images? Otherwise, the project itself has no version requirements for its dependencies, so they should always be up-to-date.

.github/workflows/ci.yml
```diff
@@ -76,12 +76,23 @@ jobs:
...
+        uses: docker/setup-buildx-action@v1
+      - name: Inspect builder
+        run: |
+          echo "Name:      ${{ steps.buildx.outputs.name }}"
+          echo "Endpoint:  ${{ steps.buildx.outputs.endpoint }}"
+          echo "Status:    ${{ steps.buildx.outputs.status }}"
+          echo "Flags:     ${{ steps.buildx.outputs.flags }}"
+          echo "Platforms: ${{ steps.buildx.outputs.platforms }}"
+      - name: Available platforms
+        run: echo ${{ steps.buildx.outputs.platforms }}
```
vincentbernat, line 90: This is a duplicate, the information is available in the previous step.

vincentbernat (changes requested):
Thanks! Looks good. See comments in the review.

I didn't enable all the supported platforms because it takes additional time to build. I don't really mind, but did you need one of the arch I didn't put?

PR #475, opened 2021-08-29
Add support for 2.5G, 5G, 25G, and 50G based Ethernet

quwer2:
The current implementation lacks support for 2.5G, 5G, 25G and 50G Ethernet.
Because of this `lldpcli show interfaces details` shows _unknown_ for _MAU oper type_ on such interfaces.

To solve this, I have extended lldpd in the follwing way:
1. Added new mautypes in `src/lib/atoms/port.c `and `src/lldp-const.h` according to [IANA-MAU-MIB](https://www.iana.org/assignments/ianamau-mib/ianamau-mib).
2. Added new cases in `src/daemon/interfaces-linux.c` to identify the new Ethernet speeds.

For 25G, I suggest to use 25GBASE-T for regular RJ45, 25GBASE-LR for fibre, and 25GBASE-CR for DAC-cables.
For 50G there is no RJ45-based standard available (yet), therefore we only have 50GBASE-LR for fibre and 50GBASE-CR for DAC-cable.
Both, 2.5G and 5G, build on RJ45 cabling, therefore 2.5GBASE-T and 5GBASE-T, respectively.


vincentbernat:
This looks good. Thanks!

quwer2:
Thanks!
The second commit 87d7bd7 adds support for the available RJ45 10GBASE-T and updates the older 100GBASE-CR10 ("phase-out" cf. [Wikipedia 100G Ethernet](https://en.wikipedia.org/wiki/100_Gigabit_Ethernet)) to the newer ("current") 100GBASE-CR4 standard.

vincentbernat:
Looks like the 10G change also needs to update some of the tests.

vincentbernat:
Thanks!

PR #478, opened 2021-09-13
Fix illumos name typo in comment

bahamat:
`illumos` instead of `Illuminos`. (Also note that the name is stylized lower case, and should almost never be capitalized.)


vincentbernat:
Thanks!

PR #481, opened 2021-09-21
lldpcli: remove redundant "ports" parameters from cap and mgmt cmds

yar-fed:
The commands below were introduced as global commands, but all of them
except the first one were missing code to ignore "ports" parameter
```
configure lldp capabilities-advertisements
unconfigure lldp capabilities-advertisements
unconfigure lldp management-addresses-advertisements
configure lldp management-addresses-advertisements
```

Signed-off-by: Yaroslav Fedoriachenko <yar.fed99@gmail.com>


vincentbernat:
Thanks!

Issue #490, opened 2021-12-09
When removing a port and immediately add it back we will receive this warning: unable to send packet on real device for Ethernet4: No such device or address 


tomer-israel:
# Bug description

when we remove a port and add it back immediately - lldpd is keep failing and cannot recover.
warning message keeps appearing:

```
[WARN/lldp] unable to send packet on real device for Ethernet4: No such device or address
```

the socket keeps failing when trying to send lldp packets.
this problem is related to the timer that aggregate events from netlink and only 1 second after the last event the ports will be updated:
[event.c lines 738-756](https://github.com/lldpd/lldpd/blob/690e447ea0a9cd488b953b95b3f1c86c8a526e03/src/daemon/event.c#L738)

if we will remove the port and immediately add the port back the deletion of the port will not be executed. 
when we remove a port and add it back we have a new ifindex and the socket that we used for this port is no longer relevant since we have a new ifindex for this port.
in these cases we need to recreate the socket with the new ifindex.

## Steps to reproduce the problem

remove a port and immediately add the port back, for example:

## Expected outcome
 no warning on log and lldp packets will be sent properly

## Current outcome

# Additional information

PR #491, opened 2021-12-09
Fix issue #490 - when receiving del event we will execute it immediately (without timer)

tomer-israel:
Signed-off-by: tomeri <tomeri@nvidia.com>

this PR will fix this issue:
https://github.com/lldpd/lldpd/issues/490

in order to recreate the socket when we remove a port and add it immediately, we can execute the delete action immediately without using the timer.

another way to solve it is to check [here](https://github.com/lldpd/lldpd/blob/690e447ea0a9cd488b953b95b3f1c86c8a526e03/src/daemon/lldpd.c#L150) if the new ifindex is equal to the current ifindex - if they are different we need to recreate the socket (levent_hardware_release & levent_hardware_init)

### DIFF
--- a/src/daemon/event.c
+++ b/src/daemon/event.c
@@ -736,8 +736,13 @@ levent_iface_recv(evutil_socket_t fd, short what, void *arg)
 	}
 
 	/* Schedule local port update. We don't run it right away because we may
-	 * receive a batch of events like this. */
-	struct timeval one_sec = {1, 0};
+	 * receive a batch of events like this.
+	 * if it's del link event we need to execute it right away in order to avoid cases where we have del event and create event in the same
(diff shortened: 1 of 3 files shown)


vincentbernat:
Unfortunately, Netlink messages can be lost and batched and therefore, it would be better to implement your second option as this one may trigger some edge cases where the bug still happens in some cases. I need to look more in the details if there is not an easy fix around the line you have highlighted or if we need to teach lldpd to use interface indexes instead of interface names to search for existing interfaces. Using names was not a good idea, but at the time, I didn't know that indexes cannot be recycled so I thought that names were more robust.

vincentbernat:
Closing. Look at #492 instead.

PR #497, opened 2022-01-11
daemon: add missing syscalls to SECCOMP filter

dawud:
When running in the foreground, additional syscalls are required


vincentbernat:
Thanks!

PR #503, opened 2022-03-09
Use BSD's upstream if_ether.h

lamaral:
Hello,

While trying to build lldpd on FreeBSD 14-CURRENT, I ran into an issue where a struct was being redefined:
```
--- liblldpd_la-interfaces-bsd.lo ---
In file included from interfaces-bsd.c:33:
In file included from /usr/include/net/if_bridgevar.h:84:
/usr/include/net/ethernet.h:62:8: fatal error: redefinition of 'ether_header'
struct ether_header {
       ^
../../include/netinet/if_ether.h:72:8: note: previous definition is here
struct  ether_header {
        ^
1 error generated.
*** [liblldpd_la-interfaces-bsd.lo] Error code 1
```
The build was working on FreeBSD 13.0-RELEASE, but started failing on CURRENT. 

I was able to pinpoint it to FreeBSD including `net/ethernet.h` on [this commit](https://github.com/freebsd/freebsd-src/commit/e9167358e402452bf83d12ac685e845099343a42).

I included `sys/socket.h` in order to have the sockaddr struct defined and moved `netinet/if_ether.h` after `netinet/in.h` in order to have `in_addr` defined before it's used in `if_ether.h`.

With these changes, I was able to build it on FreeBSD 13.0 as well as on CURRENT. Let me know what you think.

Thanks,
Luiz.


vincentbernat:
Unfortunately, this header is also used by other platforms. All the headers are a bit of a mess. They were definitely needed for older versions of Linux, but maybe this is not the case anymore. I am trying to add CI for BSD in #504. After that, I'll try to check if all these headers are still needed on Linux and MacOS. I don't think the BSD ever needed them.

Alternatively, the headers could be included only on Linux and MacOS.

vincentbernat:
So, I was able to remove non-OS specific headers. I have applied your patch and made a few tweaks. It works on OpenBSD and FreeBSD, as well as old Linux distros.

lamaral:
Awesome :)
Thanks for the fix

Issue #498, opened 2022-01-24
RFC: Inventory information fallback


harshalgohel:
#### TL;DR
If `/sys/class/dmi/` does not exist (in case of Non-EFI bootloaders), then there could be some kind of fallback mechanism from which inventory information can be fetched.
I have mentioned some approaches for this, 

**What is community opinion on these?**

---

I can think of following approaches,

1. Fallback to linux environment variables
2. Configurable via `lldpcli` - hence configurable via `lldpd.conf`
3. Specify fallback directory 

---

### Fallback to environment variables,
1. Find inventory information from EEPROM/embedded flash
2. Write `/tmp/lldpd.env` or some other path.
3. Specify `EnvironmentFile=/tmp/lldpd.env` as part of systemd service. 
4. Export variables with `export $(cat /tmp/lldpd.env)` before executing lldpd for init.d and other init services.

### Each variable configurable via `lldpcli`
After `lldpd` is started,
1. Find inventory information from EEPROM/embedded flash
2. Execute `lldpcli configure inventory <inventory_var> <value>` for each of inventory variable such as `product_serial`, `product_name`, etc.

Before `lldpd` is started, 
1. Find inventory information from EEPROM/embedded flash
2. Write/append `configure inventory <inventory_var> <value>` to `lldpd.conf` 

### Fallback directory
1. Find inventory information from EEPROM/embedded flash
2. Write `/tmp/inventory/product_serial`, `/tmp/inventory/product_name`, etc.
3. Specify fallback directory with `lldpd.conf` or `lldpcli configure med fallback-directory /tmp/inventory/`
4. Specify fallback directory as compile time flag `./configure --fallback_dmi=/tmp/inventory`

vincentbernat:
I would prefer only falling back to lldpcli and not adding yet another configuration mechanism.

PR #508, opened 2022-04-11
med: Add configure inventory commands

harshalgohel:
Default inventory information is fetched from dmi table.
Most of embedded devices do not run EFI bootloader hence dmi information
will not be available in /sys/class/dmi

- Add `configure inventory` commands
- Add `unconfigure inventory` commands
- New code is dependent on enable-lldpmed
- Add write support on lldpctl atoms
- Add support for setting inventory configuration in daemon
- Add configinventory integration tests

Github: #498

### DIFF
--- a/src/daemon/client.c
+++ b/src/daemon/client.c
@@ -18,6 +18,10 @@
 #include "lldpd.h"
 #include "trace.h"
 
+#ifdef ENABLE_LLDPMED
+#include <sys/utsname.h>
+#endif
+
 static ssize_t
 client_handle_none(struct lldpd *cfg, enum hmsg_type *type,
     void *input, int input_len, void **output, int *subscribed)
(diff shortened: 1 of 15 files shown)


vincentbernat:
Thanks! I am a bit busy, but I'll have a look this or next week. This looks good and it's great to have added tests. If you feel it, you could also update the manual page for lldpcli.

src/daemon/client.c
```diff
@@ -273,6 +277,82 @@ client_handle_get_interfaces(struct lldpd *cfg, enum hmsg_type *type,
...
+	// Follows lldpd.c - only set sw if advertising is enabled
+	if (cfg->g_config.c_advertise_version) {
+		free(local_chassis->c_med_sw);
+
+		if (!chassis->c_med_sw) {
+			if (uname(&un) < 0) {
+				log_warn("rpc", "Could not get default uname. Will continue anyway.");
+				local_chassis->c_med_sw = NULL;
+			}
+			else
```
vincentbernat, line 314: Put else on the same line as the closing brace.

src/daemon/client.c
```diff
@@ -273,6 +277,82 @@ client_handle_get_interfaces(struct lldpd *cfg, enum hmsg_type *type,
...
+				local_chassis->c_med_sw = strdup(un.release);
+		}
+		else
+			local_chassis->c_med_sw = strdup(chassis->c_med_sw);
```
vincentbernat, line 318: I think this change will get erased at regular interval if you dojn't prevent that in `lldpd.c`.

vincentbernat (changes requested):
Looks good. Tell me if you plan to write the manual page or I'll do it.

harshalgohel:
> Looks good. Tell me if you plan to write the manual page or I'll do it.

Yes I will be making changes to to lldpcli.8.in

src/daemon/client.c
harshalgohel: How do you prefer to achieve this?

I can think of a flag to check if it was manually set, if so, don't update

Why is software revision updated periodically in the first place? Not like sw will change without rebooting whatever machine it is running on....
- could kernel live-patching change software revision without reboot?

vincentbernat: Not updating software revision periodically would be the simpler workaround I think.

harshalgohel: How about moving this code structure to, https://github.com/lldpd/lldpd/blob/45f3c77332be56ce2df3bcf8c402bd1f14d29f39/src/daemon/lldpd.c#L1159
so it is called only once?

vincentbernat: Yes, this seems a good idea.

src/daemon/client.c
harshalgohel: Fixed in latest update.

harshalgohel:
> Looks good. Tell me if you plan to write the manual page or I'll do it.

Just pushed new update, with lldpcli documentation

vincentbernat (approved):

vincentbernat:
Thanks!

PR #520, opened 2022-07-04
lldp: don't send vlan when without_vlans is set

ffourcot:
Fixes: 8e9dbbf80918 "lldp: don't send VLANs when there are too many of them

On interface with a lot of vlans, we still have "Too big packet"
notification on version 1.0.14

It looks like retry does exactly the same packet than first try


vincentbernat:
Thanks. I am almost sure the additional test didn't work before the patch, but yes, the patch does nothing. I'll investigate a bit to understand where I failed.

vincentbernat:
Thanks, the test was incorrect! :-/

PR #539, opened 2022-08-17
build: ship universal binaries for macOS

vincentbernat:
This mostly reverts commit 737afb3517d822ae50722510993acbd097bc5aca.


vincentbernat:
Merged without the universal bits. When GitHub supports ARM64 runners, we should be able to ship universal binaries.

PR #551, opened 2022-10-09
oss-fuzz and fuzzer files

0x34d:
Moving fuzzing files into the main project for better harness reliability.

Signed-off-by: 0x34d <ajsinghyadav00@gmail.com>


vincentbernat:
I prefer not to mix Apache 2 with ISC. This duplicates the code present in `test/decode.c`. You can use `--enable-sanitizers=fuzzer` to not alter the configuration system.

0x34d:
> This duplicates the code present in `test/decode.c`. 

Well, it's possible. 
But eventually, this will become a mess for the linker.

0x34d:
>  You can use `--enable-sanitizers=fuzzer` to not alter the configuration system.

This flag will break the compilation of binary with the `main` function. 
because the `-fsanitize=fuzzer` library also has a `main` function.

While we are using multiple types of fuzzer for fuzzing.
I have to be specific to the linker about the fuzzing engine library.

Local Fuzzing linker for `FuzzDecode.c`:
```
LibFuzzer=-fsanitize=fuzzer 
AFL++=FuzzingEngine.a
honggfuzz=libhfuzz.a
```

oss-fuzz fuzzing linker for `FuzzDecode.c`:
```
LibFuzzer=-fsanitize=fuzzer 
AFL++=libFuzzingEngine.a
honggfuzz=libFuzzingEngine.a
```

As per your concern about the configuration system.
Now I will use a different approach.

`--enable-fuzzer` will provide fuzzing support.

For the linker, `LIB_FUZZING_ENGINE` needs to be `export LIB_FUZZING_ENGINE=fuzzing.a`,
as per the corresponding fuzzer.

So `configure.ac` can read the linker for binary.

Regards Arjun.

vincentbernat:
I was thinking about something like this: https://github.com/lldpd/lldpd/commit/9fd7b25f984569378f60ecfbe37372eb3417efd7.

But if it's too much a hurdle, we can duplicate the code.

0x34d:
> I was thinking about something like this: [9fd7b25](https://github.com/lldpd/lldpd/commit/9fd7b25f984569378f60ecfbe37372eb3417efd7).
> 
> But if it's too much a hurdle, we can duplicate the code.


Yes, This can work with a few changes in macros.

0x34d:
https://github.com/lldpd/lldpd/commit/9fd7b25f984569378f60ecfbe37372eb3417efd7#diff-7238ff164833857def0655f4fe7742553e6e42bae79089c741a599bf8aa0b02dR76

Are you sure about this one:
```
	if (cdp_decode(NULL, frame, size, hardware, &nchassis, &nport) == -1) {
```

I believe this is the right way:
```
	if (cdp_decode(NULL, frame, size, hardware, nchassis, nport) == -1) {
```

0x34d:
You can review the PR.

vincentbernat:
Sorry for the late answer. I am OK with the code change, the changes to `configure.ac` and `Makefile.am`. However, I am a bit more worried with the script and the ZIP file. The script could go in the Makefile.

Also, from a style point of view, I prefer `fuzz-decode` to `FuzzDecode`. Everything else is using snake case, not camel case.

vincentbernat:
Thanks. I have renamed and simplified a bit the build script.

PR #584, opened 2023-05-27
[Fuzzing] updates for oss-fuzz Ideal integration

0x34d:
- Separate fuzzing targets for better coverage of individual functions.
- Add more targets.
- Disable checksum while fuzzing EDP protocol to remove fuzzer roadblocks.
- Fix a memory leak in `edp_decode` .


configure.ac
```diff
@@ -154,7 +154,7 @@ AC_ARG_ENABLE([fuzzer],
   [
 case "$enableval" in
     no) fuzzer= ;;
-    yes) fuzzer="-fsanitize=fuzzer" ;;
+    yes) fuzzer="$LIB_FUZZING_ENGINE" ;;
```
0x34d, line 157: You might wonder why use `$LIB_FUZZING_ENGINE` instead of `-fsanitize=fuzzer`.

- Fuzzing is not synonymous with libFuzzer.
- Using this approach makes it easier to utilize different fuzzers with the same code. For example: `centipede`, `honggfuzz`, `AFL++`, and libFuzzer.

vincentbernat: In this case, add alternate values to yes/no (like this is done for sanitizers.

0x34d: Oh, alright.

src/daemon/protocols/edp.c
```diff
@@ -305,12 +305,14 @@ edp_decode(struct lldpd *cfg, char *frame, int s, struct lldpd_hardware *hardwar
...
 	PEEK_BYTES(chassis->c_id, ETHER_ADDR_LEN);
 
+#  ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
 	/* Let's check checksum */
 	if (frame_checksum(pos_edp, edp_len, 0) != 0) {
 		log_warnx("edp", "incorrect EDP checksum for frame received on %s",
 		    hardware->h_ifname);
 		goto malformed;
 	}
+#  endif
```
0x34d, line 315: Disable Checksum while fuzzing.

src/daemon/protocols/edp.c
```diff
@@ -345,6 +347,7 @@ edp_decode(struct lldpd *cfg, char *frame, int s, struct lldpd_hardware *hardwar
 			port->p_id_subtype = LLDP_PORTID_SUBTYPE_IFNAME;
 			edp_slot = PEEK_UINT16;
 			edp_port = PEEK_UINT16;
+			free(port->p_id);
```
0x34d, line 350: Fix a memory leak.
A memory leak occurs when `case EDP_TLV_INFO` is called twice.

Scenario:
1- Call occurs.
-----> Memory leak.
2- Call occurs.
-----> Memory is freed by the program.

Therefore, if there are two or more calls to `case EDP_TLV_INFO`, all of them leak memory, except the last one.
This change handles all memory leaks.

src/daemon/protocols/edp.c
```diff
@@ -353,6 +356,7 @@ edp_decode(struct lldpd *cfg, char *frame, int s, struct lldpd_hardware *hardwar
 				goto malformed;
 			}
 			port->p_id_len = strlen(port->p_id);
+			free(port->p_descr);
```
0x34d, line 359: Fix a memory leak.
A memory leak occurs when `case EDP_TLV_INFO` is called twice.

Scenario:
1- Call occurs.
-----> Memory leak.
2- Call occurs.
-----> Memory is freed by the program.

Therefore, if there are two or more calls to `case EDP_TLV_INFO`, all of them leak memory, except the last one.
This change handles all memory leaks.

src/daemon/protocols/edp.c
```diff
@@ -363,6 +367,7 @@ edp_decode(struct lldpd *cfg, char *frame, int s, struct lldpd_hardware *hardwar
 			PEEK_DISCARD_UINT16; /* vchassis */
 			PEEK_DISCARD(6);     /* Reserved */
 			PEEK_BYTES(version, 4);
+			free(chassis->c_descr);
```
0x34d, line 370: Fix a memory leak.
Fix a memory leak.
A memory leak occurs when `case EDP_TLV_INFO` is called twice.

Scenario:
1- Call occurs.
-----> Memory leak.
2- Call occurs.
-----> Memory is freed by the program.

Therefore, if there are two or more calls to `case EDP_TLV_INFO`, all of them leak memory, except the last one.
This change handles all memory leaks.

README.md
```diff
@@ -358,28 +358,56 @@ To enable code coverage, use:
...
-that:
+build and run:
+```
+./configure --disable-shared --enable-pie --enable-fuzzer=$LIB_FUZZING_ENGINE
+make
+cd tests/
+
+./fuzz_bitmap 
+./fuzz_fixedpoint 
+./fuzz_marshal
```
0x34d, line 392: I will remove this in new PUSH.

src/daemon/protocols/edp.c
0x34d, line 370: Fix a memory leak.
A memory leak occurs when case `EDP_TLV_INFO` is called twice.

Scenario:
1- Call occurs.
-----> Memory leak.
2- Call occurs.
-----> Memory is freed by the program.

Therefore, if there are two or more calls to case `EDP_TLV_INFO`, all of them leak memory, except the last one.
This change handles all memory leaks.

PR #596, opened 2023-08-07
[bug] Use-of-uninitialized-value, Integer-overflow

0x34d:
bug : [Use-of-uninitialized-value](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=59532&q=proj%3Dlldpd&can=2) [Integer-overflow](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60046&q=proj%3Dlldpd&can=2)


0x34d:
@vincentbernat Ping

PR #597, opened 2023-08-14
Modify custom TLV information to be string instead of integers

Shal-sudo:
Update OUI_INFO type to be char *
Update display and conf-lldp portions to reflect the changes

Tested locally

cc: @vincentbernat


vincentbernat:
Ability to pass arbitrary bytes is still needed for many custom TLV.

Shal-sudo:
Fair. I will change this PR to add a new command that allows string field for custom tlv
eg: configure custom-tlv lldp oui 11,22,33 subtype 44 oui-info-str "hello"

PR #621, opened 2023-11-24
client: translate dot into the short hostname

gsauthof:
as documented in lldpcli(8):

    configure system hostname name

        Override system hostname with the provided value. By default, the
        system name is the FQDN found from the resolved value of uname
        -n.  As a special value, use "." (dot) to use the short hostname
        instead of a FQDN.


vincentbernat:
Thanks!

PR #622, opened 2023-11-24
build: improve const correctness and prototype declarations

gsauthof:
String literals aren't assigned to non-const char pointers anymore, they aren't returned via non-const char pointers, const strings aren't passed arround via non-const pointers etc.

As a side effect, this eliminates all -Wdiscarded-qualifiers warnings.

Symbol visibility is minimized by declaring additional functions static, i.e those which are only used inside a single translation unit.

Also, a few includes are added such that all non-static function definitions 'see' their declarations - such that accidental differences in signatures immedialy yield compile errors or warnings.

A few macros are also modified to emit forward declarations for non-static functions in order to eliminate noise when compiling with -Wmissing-prototypes.
However, it might make sense to move those non-static function definitions from a header into a dedicated translation unit, at some point.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -90,6 +90,7 @@ AX_CFLAGS_GCC_OPTION([-Wheader-guard], [LLDP_CFLAGS])
 AX_CFLAGS_GCC_OPTION([-Wdocumentation], [LLDP_CFLAGS])
 AX_CFLAGS_GCC_OPTION([-Winline], [LLDP_CFLAGS])
 AX_CFLAGS_GCC_OPTION([-Wpointer-arith], [LLDP_CFLAGS])
+AX_CFLAGS_GCC_OPTION([-Wmissing-prototypes], [LLDP_CFLAGS])
 AX_CFLAGS_GCC_OPTION([-Wno-cast-align], [LLDP_CFLAGS]) dnl clang is bad at this
 AX_CFLAGS_GCC_OPTION([-Wno-unused-parameter], [LLDP_CFLAGS])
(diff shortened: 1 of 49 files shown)


vincentbernat:
Looks good to me. Maybe you want to enable the warning by default so that we don't introduce regressions in the future (CI runs with `-Werror`)?

Also, please run clang-format if you can.

gsauthof:
Sure, I've updated the pull request.

However, that clang-format normalizes

```
if (c)
    *c = 0;
```

to

```
if (c) *c = 0;
```

made me cringe a little. :-)

---

I've also added `-Wmissing-prototypes` to configure.ac.

The `-Wdiscarded-qualifiers` options should be enabled, by default.
And gcc man page only documents `-Wno-discarded-qualifiers`.

Thus, I don't quite understand why your CI builds didn't fail before because you are compiling with `-Werror`  there.

I see how you are setting `-Werror` in `tests/ci/run.sh` and when I do it the same way locally it's simply appended to the default CFLAGS, as expected. And since the default CFLAGS already contain `-Wall` and more we should see `-Wdiscarded-qualifiers` errors in CI builds but we don't?

vincentbernat:
Use of `-Werror` was broken in the CI script (it was invoking `make -Werror` on Linux instead of `make CFLAGS=-Werror`). I have pushed a commit to fix that. If you rebase, you should see the errors.

vincentbernat:
Ready to merge?

gsauthof:
Yes.

It just required a few more force pushes due to different configure options exposing additional code paths and test programs.

vincentbernat:
Thanks!

PR #635, opened 2024-02-08
docs: man page fixes

systemcrash:
changed: network cable -> link

change -M help text in src/daemon/lldpd.c also

Please verify the explanations under the following (where ambiguous ... not ... and ... was removed )
```
system chassisid
system capabilities enabled
```


vincentbernat:
Yes, it is correct. Thanks!

PR #649, opened 2024-04-11
lib: Fix a buffer underflow in _lldpctl_dump_in_atom

clifclar:
Do not write a zero byte before the beginning of the buffer when _lldpctl_dump_in_atom is called with a zero length atom.

I ran SonarQube and it found this issue.  I'm not sure if it is valid to call the dump function on a zero-length atom or not though.


vincentbernat:
Thanks!

PR #656, opened 2024-04-25
Fix macos warnings at compile time.

systemcrash:
If this happens:

lldpd.c:1577:14: warning: variable 'pidfile' set but not used [-Wunused-but-set-variable]
        const char *pidfile = LLDPD_PID_FILE;

Then the other places using pidfile should get the same ifndef treatment


vincentbernat:
Thanks!

PR #661, opened 2024-05-11
priv: disable LLDP in firmware for Intel X7xx cards on FreeBSD

### LINKED ISSUE #660: priv does not disable Intel X7xx firmware LLDP on FreeBSD
# Bug description

a74fd8c only disables the Intel 700 series firmware LLDP process on Linux. See also: 45f3c77

## Steps to reproduce the problem

 1. Install the `net-mgmt/lldpd` port on a FreeBSD machine equipped with an Intel X7xx NIC (see `ixl(4)`)
 2. Enable the `lldpd` service (e.g. `sysrc lldpd_enable=YES`)
 3. Start the `lldpd` service (e.g. `service lldpd start`)
 4. Run `sysctl dev.ixl | grep lldp`
 6. Observe that the value of each `dev.ixl.N.fw_lldp` sysctl is 1.

## Expected outcome

`lldpd` has disabled the `ixl` firmware LLDP process.

## Current outcome

`lldpd` leaves the `ixl` firmware LLDP enabled.

# Additional information

Note that this is from a local ports tree where the `net-mgmt/lldpd` update from 1.0.14 to 1.0.18 has not yet been submitted/accepted upstream.

 - Output of `lldpd -vv`:

```
lldpd 1.0.18
  Built on 2024-04-10T16:46:36Z

Additional LLDP features:    LLDP-MED, Dot1, Dot3, Custom TLV
Additional protocols:        CDP, FDP, EDP, SONMP
SNMP support:                yes
Privilege separation:        enabled
Privilege separation user:   _lldpd
Privilege separation group:  _lldpd
Privilege separation chroot: /var/run/lldpd
Configuration directory:     /usr/local/etc

C compiler command: cc  -Wunknown-warning-option -fdiagnostics-show-option -fdiagnostics-color=auto -fno-omit-frame-pointer -funsigned-char -pipe -Wall -W -Wextra -Wformat -Wformat-security -Wimplicit-fallthrough -Wfatal-errors -Wheader-guard -Wdocumentation -Winline -Wpointer-arith -Wmissing-prototypes -Wno-cast-align -Wno-unused-parameter -Wno-missing-field-initializers -Wno-sign-compare -fstack-protector -fstack-protector-all -fstack-protector-strong -fstack-clash-protection  -D_FORTIFY_SOURCE=2 -D_GNU_SOURCE -O2 -pipe  -fstack-protector-strong -isystem /usr/local/include -fno-strict-aliasing -std=gnu99  -isystem /usr/local/include
Linker command:     /usr/bin/ld  -Wl,-z,relro -Wl,-z,now   -Wl,-rpath,/usr/local/lib -fstack-protector-strong  -L/usr/local/lib
```

 - Output of `uname -sro`:

```
FreeBSD 13.3-RELEASE
```

neirbowj:
Fixes #660.

Tested on FreeBSD 13.3-RELEASE (amd64) with an Intel X710 running firmware version "fw 8.84.66032 api 1.14 nvm 8.40 etid 8000c815 oem 20.5120.11".

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,7 @@ lldpd (1.0.19)
  * Changes:
    + Support of both Apple Silicon and Intel for macOS package.
    + Add cvlan/svlan/tpmr capabilities.
+   + Disable LLDP in firmware for Intel X7xx cards on FreeBSD.
 
 lldpd (1.0.18)
  * Changes (breaking):
--- a/README.md
+++ b/README.md
@@ -449,8 +449,15 @@ for f in /sys/kernel/debug/i40e/*/command; do
     echo lldp stop > $f
 done
 ```
+On FreeBSD, use `sysctl` stop the embedded LLDP daemon:
+```sh
+for oid in $(sysctl -Nq dev.ixl | grep fw_lldp); do
+    sysctl $oid=0
+done
+```
 This may also apply to the `ice` (Intel E8xx cards) driver. These
-steps are not necessary with a recent version of `lldpd` (1.0.11+).
+steps are not necessary with a recent version of `lldpd` (1.0.11+ for
+Linux, 1.0.19+ for FreeBSD).
 
 ## License
 
--- a/src/daemon/priv-bsd.c
+++ b/src/daemon/priv-bsd.c
@@ -17,14 +17,81 @@
 
 #include "lldpd.h"
 
+#include <ctype.h>
 #include <unistd.h>
+#include <net/if.h>
 #include <net/bpf.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/sysctl.h>
 #include <sys/ioctl.h>
 #include <fcntl.h>
 #include <errno.h>
 #include <string.h>
+
+#if defined HOST_OS_FREEBSD
+#  include <net/if_mib.h>
+
+/* Quirks needed by some additional interfaces. Currently, this is limited to
+ * disabling LLDP firmware for ixl. OpenBSD and NetBSD disable the ixl
+ * firmware LLDP agent upon driver attach. */
+static void
+asroot_iface_init_quirks(int ifindex, char *name)
+{
+	int sysctl_oid[6];
+	int fw_lldp = 0;
+	size_t driver_name_len;
+	char *driver_name = NULL;
+	char *sysctl_name = NULL;
+
+	/* Check driver. */
+	sysctl_oid[0] = CTL_NET;
+	sysctl_oid[1] = PF_LINK;
+	sysctl_oid[2] = NETLINK_GENERIC;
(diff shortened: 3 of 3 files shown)


neirbowj:
For posterity's sake, here is the basis for my comment about [OpenBSD](https://github.com/openbsd/src/blob/c4602437ad9d125ac3da930b539a6fc847740b2b/sys/dev/pci/if_ixl.c#L1857) and [NetBSD](https://github.com/NetBSD/src/blob/abfde61b9201ed4c3cb4e5266c760d79f2b51a98/sys/dev/pci/if_ixl.c#L1212).

macOS does not define the `IFDATA_DRIVERNAME` OID in `if_mib.h`. If it is not possible to rename an interface, it would not be needed, and one could reliably detect the `ixl` driver and derive the `fw_lldp` OID name from the interface name. In any case, I have no way to test this on a macOS machine.

vincentbernat:
Thanks!

neirbowj:
For general awareness, I have discovered what I suspect may be a bug in the interaction between lldpd and the X710 described above, now running FreeBSD 14.1-RELEASE-p2. As far as I can tell, this patch does not introduce nor trigger the bug. I am leaving this comment for those who may find this while troubleshooting the same symptoms. I will follow up if I can learn more.

If the following conditions hold, then the IPv6 multicast MAC (33:33:xx:xx:xx:xx) filter will fail to be correctly programmed into the X710s, and the system will not receive the required IPv6 multicast traffic via the LAGG (such as Neighbor Discovery messages).

* lldpd 1.0.14 enabled to start at boot
* a pair of X710 NICs are configured as an LACP [LAGG](https://man.freebsd.org/cgi/man.cgi?query=lagg&apropos=0&sektion=0&manpath=FreeBSD+14.1-RELEASE+and+Ports&arch=default&format=html)
* a VLAN is tagged on the LAGG
* the [VLAN interface](https://man.freebsd.org/cgi/man.cgi?query=vlan&apropos=0&sektion=0&manpath=FreeBSD+14.1-RELEASE+and+Ports&arch=default&format=html) is added as a member of a [bridge interface](https://man.freebsd.org/cgi/man.cgi?query=bridge&apropos=0&sektion=0&manpath=FreeBSD+14.1-RELEASE+and+Ports&arch=default&format=html)
* one half of an [epair interface](https://man.freebsd.org/cgi/man.cgi?query=epair&apropos=0&sektion=0&manpath=FreeBSD+14.1-RELEASE+and+Ports&arch=default&format=html) is also added to the bridge
* IPv6 is enabled on the other end of the epair

Some of these conditions may be optional. On a substantially similar system, where the main difference in the network configuration is that there is no bridge or epair (i.e., IPv6 is configured directly on the VLAN interface), the problem does not occur. Removing the epair from the bridge and re-adding it restores normal IPv6 functionality.

PR #666, opened 2024-06-06
Avoid a NULL pointer deref when reading os-release

ngie-eign:


vincentbernat:
Ack for the code change, but for `.gitignore`, you can avoid that by using an out-of-build tree:

ngie-eign:
> Ack for the code change, but for `.gitignore`, you can avoid that by using an out-of-build tree:
> 
> ```
> mkdir build
> cd build
> ../configure ...
> ```
> 
> This reduces the maintenance burden on `.gitignore`.

I suppose so. I've gotten used to projects not fully supporting that out of the box.

vincentbernat:
Thanks!

ngie-eign:
Ughhh.. I probably should have added a NULL check for the next strtok call 🤦. I’ll submit a patch for that too.

PR #672, opened 2024-07-10
add range restriction for tx hold and interval

liuhangbin:
Based on IEEE 802.1AB(2016) 9.2.5. Add a valid range for tx hold and interval.
Also set the limit of tx ttl.


vincentbernat:
Thanks!

PR #681, opened 2024-10-15
fix endless loop in redefined malloc/remalloc functions on system usi…

lzhouAlan:
lldpcli and lldpd will entere into endless loop on our hardware box after upgrading lldap from 1.0.4 to 1.0.18. 
This issue is caused by endless call in malloc/remalloc function.

### DIFF
--- a/src/compat/malloc.c
+++ b/src/compat/malloc.c
@@ -1,10 +1,10 @@
 /* -*- mode: c; c-file-style: "openbsd" -*- */
 /* malloc replacement that can allocate 0 byte */
 
-#undef malloc
 #include <stdlib.h>
 #include <sys/types.h>
 #include "compat.h"
+#undef malloc
 
 /* Allocate an N-byte block of memory from the heap.
    If N is zero, allocate a 1-byte block.  */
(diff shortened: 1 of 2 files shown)


lzhouAlan:
undef malloc brefore compiling rpl_malloc, or else it will be translated as follows:
void *
rpl_malloc(size_t n)
{
    if (n == 0) n = 1;
    return **malloc**(n);
}

===>
void *
rpl_malloc(size_t n)
{
    if (n == 0) n = 1;
    return **rpl_malloc**(n);
}

lzhouAlan:
undef malloc a compiling rpl_malloc, or else it will be translated as follows:
void *
rpl_malloc(size_t n)
{
    if (n == 0) n = 1;
    return **malloc**(n);
}

===>
void *
rpl_malloc(size_t n)
{
    if (n == 0) n = 1;
    return **rpl_malloc**(n);
}

lzhouAlan:
@vincentbernat  It seems that CI/build-linux check failed. What should I do next ?

vincentbernat:
Rebase your change on master, it should be fixed now.

lzhouAlan:
> Rebase your change on master, it should be fixed now.

Rebased

lzhouAlan:
@vincentbernat  failed again

Issue #680, opened 2024-10-14
TX Fast Start only enabled when a MED peer is detected


rbu9fe:
In my opinion TX fast start should always be enabled but currently it's only enabled if 
1. MED support is enabled at compile time and 
2. a new neighbor supporting MED is detected.

IEEE 802.1AB 2016 doesn't define MED at all (as it's just an extension) and TX fast start is a general LLDP property and always enabled.
So why does this implementation enable it only under the conditions above?

rbu9fe:
It was also added into IEEE 802.1AB 2009 so I believe it should generally be enabled at least when MED support is disabled either at compile-time already by not defining `ENABLE_LLDPMED` or at runtime by not enabling the MED protocol when starting the lldpd daemon.

vincentbernat:
Yes, I agree.

rbu9fe:
I added a [PR ](https://github.com/lldpd/lldpd/pull/683) with a corresponding fix.

rbu9fe:
Are you missing something for this to be accepted? I'd be glad to help.

PR #683, opened 2024-10-22
fix: Always TX fast start complient to IEEE802.1as-2009

rbu9fe:
That's a PR for [this issue](https://github.com/lldpd/lldpd/issues/680). It works for me, however, I did not adjust unit tests.


vincentbernat:
Unfortunately, without MED support, this cannot be disabled/modified. There are more modifications to be done to the CLI.

jeromesun14:
Request for this feature!

PR #689, opened 2024-10-28
fix: add socket lock to AppArmor policy

sspans:
Resolves:

```
2024-10-27T20:05:27 [WARN/lldpctl] cannot get lock on /run/lldpd/lldpd.socket.lock: Permission denied
2024-10-27T20:05:27 [INFO/lldpctl] an error occurred while executing last command
2024-10-27T20:05:27 [INFO/lldpctl] lldpd should resume operations
```


vincentbernat:
Thanks!

PR #693, opened 2024-11-11
feat: Add lldpctl C++ wrapper

rbu9fe:
In case you're interested here is a C++ wrapper that supports automatic lifetime management of the underlying resources. For C++ projects that greatly simplifies resource management.

Some features are not yet available, I just wanted to get early feedback whether you're interested in integrating that wrapper.

### DIFF
--- a/README.md
+++ b/README.md
@@ -429,7 +429,9 @@ To embed lldpd into an existing system, there are two points of entry:
     parse its output for integration and automation purpose. Another
     way is to use SNMP support. A third way is to write your own
     controller using `liblldpctl.so`. Its API is described in
-    `src/lib/lldpctl.h`. The custom binary protocol between
+    `src/lib/lldpctl.h`. For C++, there is a wrapper class in 
+    `src/lib/lldpctl.hpp` with automatic lifetime management of
+    underlying resouces. The custom binary protocol between
     `liblldpctl.so` and `lldpd` is not stable. Therefore, the library
     should always be shipped with `lldpd`. On the other hand, programs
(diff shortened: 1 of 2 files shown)


vincentbernat:
Yes, I can take it.

rbu9fe:
How can a `lldpctl_change_callback2` callback be unregistered again? By closing the underlying connection? Is that even allowed (since the doc says not to use that connection for anything else besides receiving notifications)?

vincentbernat:
You can unregister by passing NULL, but it won't remove the state allowing to use the connection for anything else. Maybe this could be reusable, but I think there may be a race condition if a notification is ongoing. Better close the connection.

rbu9fe:
The C++ wrapper is now ready. It comes with some convenience features like managing the lifetime of an atom's underlying connection or of its parent atom (if available).

An atom's value can be retrieved as a `std::string`, `std::string_view`, `int` or as a `std::span<std::byte>`. Atom lists can be get as `std::vector`.

Note: Stopping the watcher thread doesn't work yet as the underlying `lldpctl_watch` doesn't unblock, therefore, that must be fixed in the C library.

rbu9fe:
Solved stopping the watcher thread by using `lldpctl_watch_sync_unblock` from [this PR](https://github.com/lldpd/lldpd/pull/696). Therefore, that PR must be merged before this C++ wrapper.

vincentbernat:
Could you rebase your PR? Thanks.

rbu9fe:
Done.

vincentbernat:
Thanks. I'll see if I am able to keep it up-to-date.

Issue #675, opened 2024-09-17
lldpctl_watch leads to hung on exit


vlaggeor:
Writing a simple application that has

```
lldpctl_watch_callback(conn, watchcb, NULL)
...
lldpctl_watch(conn)
...
lldpctl_release(conn);
```

will lead to a hung.

This is because because `lldpctl_watch` remains in a `while` loop on `watch_triggered` that will be called only. when change is received and notification is called  in `check_for_notification` 

```
int
lldpctl_watch(lldpctl_conn_t *conn)
{
	int rc = 0;

	RESET_ERROR(conn);

	if (conn->state != CONN_STATE_WATCHING)
		return SET_ERROR(conn, LLDPCTL_ERR_INVALID_STATE);

	conn->watch_triggered = 0;
	while (!conn->watch_triggered) {
		rc = _lldpctl_needs(conn, 1);                            <-------
		if (rc < 0) return SET_ERROR(conn, rc);
	}

	RESET_ERROR(conn);
	return 0;
}
```

```
ssize_t
lldpctl_recv(lldpctl_conn_t *conn, const uint8_t *data, size_t length)
{

.....

	/* Read all notifications */
	while(!check_for_notification(conn));        <----------

	RESET_ERROR(conn);

	return conn->input_buffer_len;
}
```

```
static int
check_for_notification(lldpctl_conn_t *conn)
{
....
	rc = ctl_msg_recv_unserialized(&conn->input_buffer,
	    &conn->input_buffer_len,
	    NOTIFICATION,
	    &p,
	    &MARSHAL_INFO(lldpd_neighbor_change));
	if (rc != 0) return rc;              <----------
	change = p;

	/* We have a notification, call the callback */
	if (conn->watch_cb) {
....
		conn->watch_triggered = 1;       <---------- 
		goto end;
	}
```

vlaggeor:
A proposed fix is in https://github.com/lldpd/lldpd/pull/676

rbu9fe:
I face the same issue now. Since the `lldpctl_watch` function is `read()`-blocked it must somehow be unblocked. I don't see how this is achieved by #676.

I guess whether to unblock implicitly when closing the connection or to add a corresponding public function is just a design flavor. In any case this must be triggered by a separate thread.

@vincentbernat Any ideas how to unblock the `read()`-blocked thread? Perhaps by a sequence similar to:
1. Invalidate the connection (e.g. by setting the state to something invalid).
[...]

PR #696, opened 2024-11-28
feat: Add lldpctl_watch_sync_unblock

rbu9fe:
This PR solves #675 by adding the new function `lldpctl_watch_sync_unblock`. It's applicable only for synchronous callback connections as others have their read function under control anyway. 

The change is thread safe since `lldpctl_watch_sync_unblock` only needs to check whether the connection refers to a synchronous callback connection. It then unblock another thread's `lldpctl_watch` call by sending data using a pipe, which listens to data from  both the daemon and the pipe.

A new error code `LLDPCTL_ERR_CALLBACK_UNBLOCK` has been introduced in order to distinguish the nature of why `lldpctl_watch` failed.

### DIFF
--- a/src/client/lldpcli.c
+++ b/src/client/lldpcli.c
@@ -43,6 +43,11 @@ extern const char *__progname;
 static struct cmd_node *root = NULL;
 const char *ctlname = NULL;
 
+/* Global context required for graceful termination */
+static struct {
+	lldpctl_conn_t **conn;
+} shutdown_context;
+
 static int
 is_lldpctl(const char *name)
 {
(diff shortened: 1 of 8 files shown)


rbu9fe:
Replaced `select` by `epoll` as it doesn't suffer the file descriptor limitations that `select` has.

vincentbernat:
The change is unrelated to your initial PR, isn't it? `select()` is POSIX.2001 while `poll()` is more recent, so maybe this is the right time to switch to `poll()`. But this could be done in another PR.

vincentbernat:
It looks good to me. Could you use it with `lldpcli`? Maybe on SIGTERM, you could unblock and close properly (and quit if you don't want to reopen a connection, thought it would show how this could work).

rbu9fe:
Accidentally pushed a bad change in between but now it's ready including your proposed lldpcli changes.
I simply call `lldpctl_watch_sync_unblock` from the signal handler, which internally calls `write`, which is an _async-signal-safe_ POSIX function. When the connection is not referring to a watcher then `lldpctl_watch_sync_unblock` simply returns without doing anything so calling it is fine in any case.

vincentbernat:
Thanks. I'll add another commit to fix the versioning.

PR #704, opened 2025-01-12
autoconf: Use m4sh syntax where appropriate

Polynomial-C:
This fixes an issue where when using the options

  `--without-embedded-libevent --without-libbsd`

configure errors out with missing libevent:
```
checking for libevent >= 2.0.5... no
configure: error: *** libevent not found
```
but succeeds to find libevent when using

  `--without-embedded-libevent --with-libbsd`

configure options instead.
See also: https://autotools.info/autoconf/m4sh.html

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -45,9 +45,9 @@ gl_LD_VERSION_SCRIPT
 # Checks for programs.
 AC_PROG_CC
 AC_PROG_CC_C99
-if test x"$ac_cv_prog_cc_c99" = x"no"; then
+AS_IF([test x"$ac_cv_prog_cc_c99" = x"no"], [
   AC_MSG_FAILURE([*** C99 support is mandatory])
-fi
+])
 AC_PROG_CPP
 AM_PROG_CC_C_O
 LT_INIT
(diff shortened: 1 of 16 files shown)


Polynomial-C:
Not sure why the fuzzing CI jobs fail. I've tried to reproduce this locally but all configure calls regarding `--enable-fuzzer` succeeded.

Polynomial-C:
> 
> Are these constructs compatible with older autoconf?
> 

They should be at least compatible with `autoconf-2.69` or newer. I didn't have a chance to test any older autoconf release though (using Gentoo Linux as platform).

vincentbernat:
Does it work with autoconf 2.69 for you?

Polynomial-C:
Not sure why the fuzzing CI jobs fail. I've tried to reproduce this locally but all configure calls regarding `--enable-fuzzer` succeeded.

> Does it work with autoconf 2.69 for you?

It did. I had a typo in my changes in `configure.ac`. Let's see if fuzzing tests now no longer fail.

vincentbernat:
Thanks!

Issue #707, opened 2025-01-20
Feature Request: Provide reason why LLDP neighbor was deleted


rbu9fe:
There are various reasons why an LLDP neighbor may be deleted from the MIB:
- The neighbor signed off by sending a termination LLDP PDU (with TTL=0).
- The neighbor's LLDP PDU timed out.
- The interface was removed from the interface pattern (note that IEEE802.1AB doesn't define that, instead, it specifies to simply wait for the timeout after disabling an interface)

In our application we need to know the reason for the deletion. Currently, watch callbacks only receive the event `lldpctl_c_deleted` not knowing the exact reason behind. I'm planning to provide a PR for that feature.

Main question is whether the extension must be backwards compatible so that existing consumers of the lldpd package don't have to be adjusted. If yes then I see two options:
1. When a LLDP neighbor is deleted then send two events: `lldpctl_c_deleted` followed by a new event indicating the reason, e.g. `lldpctl_c_deleted_timeout`.
2. Introduce a new build option and enable the new feature only when that option is enabled in the build system, replacing `lldpctl_c_deleted` by the event describing the reason for the deletion. In that solution there would be `#ifdef`'s in the code that control whether `lldpctl_c_deleted` is delivered or one of the new deletion event types.

In any case `lldpctl_c_deleted` could be declared as deprecated to encourage clients to update their code to replace `lldpctl_c_deleted`.

What's your preferred way how to implement that feature?

rbu9fe:
See the referenced PR with a proposal. I would appreaciate feedback.

PR #708, opened 2025-01-20
Tx fast start neighbor deletion reason

rbu9fe:
This is the draft PR for #707.
To be discussed:
- If `lldpctl_c_deleted` event shall be kept for backward compatibility:
  - Add build parameter to build with support for either only `lldpctl_c_deleted` or for `lldpctl_c_deleted_signoff`, `lldpctl_c_deleted_timeout` and `lldpctl_c_deleted_admin`?
  - Deliver both `lldpctl_c_deleted` and the more specific event `lldpctl_c_deleted_signoff`, `lldpctl_c_deleted_timeout` or `lldpctl_c_deleted_admin`?
    - Let `lldpcli` ignore `lldpctl_c_deleted` as it expects a more specific event anyway?
  - Tag `lldpctl_c_deleted` as deprecated?
- If backward compatibility is not required then simply replace `lldpctl_c_deleted` by `lldpctl_c_deleted_signoff`, `lldpctl_c_deleted_timeout` and `lldpctl_c_deleted_admin`.

Any feedback is highly appreciated.

### DIFF
--- a/src/client/conf-lldp.c
+++ b/src/client/conf-lldp.c
@@ -23,6 +23,65 @@
 #include "../log.h"
 
 static int
+cmd_faststart(struct lldpctl_conn_t *conn, struct writer *w, struct cmd_env *env,
+    const void *arg)
+{
+	log_debug("lldpctl", "configure fast interval support");
+
+	lldpctl_atom_t *config = lldpctl_get_configuration(conn);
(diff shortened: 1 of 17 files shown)


rbu9fe:
Created from wrong source branch

PR #709, opened 2025-01-20
Provide reason why LLDP neighbor was deleted

rbu9fe:
This is the draft PR for #707.
To be discussed:
- If `lldpctl_c_deleted` event shall be kept for backward compatibility:
  - Add build parameter to build with support for either only `lldpctl_c_deleted` or for `lldpctl_c_deleted_signoff`, `lldpctl_c_deleted_timeout` and `lldpctl_c_deleted_admin`?
  - Deliver both `lldpctl_c_deleted` and the more specific event `lldpctl_c_deleted_signoff`, `lldpctl_c_deleted_timeout` or `lldpctl_c_deleted_admin`?
    - Let `lldpcli` ignore `lldpctl_c_deleted` as it expects a more specific event anyway?
  - Tag `lldpctl_c_deleted` as deprecated?
- If backward compatibility is not required then simply replace `lldpctl_c_deleted` by `lldpctl_c_deleted_signoff`, `lldpctl_c_deleted_timeout` and `lldpctl_c_deleted_admin`.

Any feedback is highly appreciated.


rbu9fe:
Somehow the merge checks above failed. How can I run those checks in advance? Using an Ubuntu VM I tried `make check` which succeeds, whereas `./tests/ci/run.sh` fails on both master and this PR with the same error:
> ...
> autogen.sh: for the next step, run ./configure
> mkdir: cannot create directory ‘build’: File exists
> ./tests/ci/run.sh: 25: ../configure: not found
> cat: config.log: No such file or directory

Any suggestions?

rbu9fe:
The merge checks for this PR failed. Using an Ubuntu VM I `make check` succeeds but `./tests/ci/run.sh` fails with
> ImportError while loading conftest '/home/ubuntu/development/github/lldpd/tests/integration/conftest.py'.
> conftest.py:22: in <module>
>     ???
> E   AttributeError: module 'pytest' has no attribute 'helpers'

I guess some prerequisite is missing to run those tests. Any suggestions? The readme doesn't really say how to run those tests.

vincentbernat:
In the CI, it shows why, for example: https://github.com/lldpd/lldpd/actions/runs/12871171217/job/35883869067?pr=709#step:5:1131

So, you need to update the test to display the reason too.

For running tests locally, you need to create a virtualenv and install `tests/integration/requirements.txt` in it. There is a hint about that in `tests/ci/install.sh`, but on GitHub, we have a one-time VM and everything is done as root.

rbu9fe:
I'll work on the tests when we have agreed on the approach how to implement that feature. What's your favourite approach [here](https://github.com/lldpd/lldpd/pull/709#issue-2799514046)?

rbu9fe:
The tests work now.

systemcrash:
Masks? so generic DELETE operations all fall under a specific category:
0x10 = delete
0x20 = add
etc
so 0x10 + deletetimeout (0x1) = 0x11
so 0x10 + deleteadmin (0x2) = 0x12

This way you can still flag the generic delete and more specific reason.

rbu9fe:
Fine for me. The main question is how to handle backwards incompatibility. Options I see are:
- Backward compatibility is required:
  - Either enable new feature only when actively activated by the user, or
  - we assume that new events are ignored by existing consumers; then keep the original event (but perhaps flag it as deprecated?) and add new ones.
- Backward compatibility is not required:
  - Simply define desired behaviour (send generic + specific deletion events, only specific event, use event mask, ...).

I'm fine with any approach as long as it gets approved and merged.

systemcrash:
So either a config flag, or, simply send both. Send the specific reason event, then the generic reason event. Is that not possible?

rbu9fe:
Performance-wise I'd like to avoid having to send two events but if that's required for backwards compatibility I'm fine.
@vincentbernat What's your opinion?

systemcrash:
Is backwards compatibility on the receiver side? If so, then it would seem the choice is obvious.

rbu9fe:
This PR doesn't change the behaviour on the wire but only on the library's API.

Issue #710, opened 2025-01-30
missing seccomp rule for clock_gettime64()


cbricart:
on a (recent) x86 (i.e 32-Bit) system, `lldpd` with seccomp enabled just terminates with `SIGSYS`.
it just works fine without seccomp disabled during build.

`strace -f lldpd -d` yields:

```
…
[pid 21924] statx(AT_FDCWD, "/etc/nsswitch.conf", AT_STATX_SYNC_AS_STAT|AT_NO_AUTOMOUNT, STATX_BASIC_STATS, {stx_mask=0, stx_attributes=0, ...}) = 383
[pid 21924] --- SIGSYS {si_signo=SIGSYS, si_code=SYS_SECCOMP, si_call_addr=0xb7f10302, si_syscall=__NR_statx, si_arch=AUDIT_ARCH_I386} ---
[pid 21924] getpid()                    = 21924
[pid 21924] clock_gettime64(CLOCK_REALTIME_COARSE, {tv_sec=0, tv_nsec=0}) = 403
[pid 21924] --- SIGSYS {si_signo=SIGSYS, si_code=SYS_SECCOMP, si_call_addr=0xb7f10302, si_syscall=__NR_clock_gettime64, si_arch=AUDIT_ARCH_I386} ---
[pid 21926] <... read resumed>"", 4)    = 0
[pid 21924] +++ killed by SIGSYS +++
[pid 21926] exit_group(0)               = ?
[pid 21926] +++ exited with 0 +++
```

so I assume `clock_gettime64` is just missing alongside `clock_gettime` in `src/daemon/priv-seccomp.c`?

vincentbernat:
Thanks!

PR #711, opened 2025-01-30
add more missing syscalls to seccomp profile

cbricart:
it turned out, adding just clock_gettime64()
from #710 has not been enough for ia32 platform.

iteratively added additional syscalls, that
prevented lldpd from starting up without
bailing out with SIGSYS


vincentbernat:
Thanks!

PR #723, opened 2025-03-25
man: Amend keyword colour formatted as operations and not values.

systemcrash:
the keywords `supported` etc should be red here:

Screenshot 2025-03-25 at 23 23 33


vincentbernat:
You need to prefix the macro with a dot (`.Sy`).

vincentbernat:
Thanks!

PR #724, opened 2025-03-26
build: Fix macos compile warning on macos >= 12

systemcrash:
dmi-osx.c:37:40: warning: 'kIOMasterPortDefault' is deprecated: first deprecated in macOS 12.0 [-Wdeprecated-declarations]
        service = IOServiceGetMatchingService(kIOMasterPortDefault, matching);
                                              ^~~~~~~~~~~~~~~~~~~~
                                              kIOMainPortDefault

### DIFF
--- a/src/daemon/dmi-osx.c
+++ b/src/daemon/dmi-osx.c
@@ -34,7 +34,8 @@ dmi_get(const char *classname, CFStringRef property)
 		    classname);
 		goto end;
 	}
-	service = IOServiceGetMatchingService(kIOMasterPortDefault, matching);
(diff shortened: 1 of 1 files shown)


vincentbernat:
The new one is not really documented and the [documentation for`IOServiceGetMatchingService`](https://developer.apple.com/documentation/iokit/1514535-ioservicegetmatchingservice) has not been updated. To avoid carrying an ifdef forever, could you use NULL instead of the only occurrence of this constant?

systemcrash:
Are you certain that it's NULL? I've seen mentioned that it's a synonym for zero (`0`)

systemcrash:
Done.

PR #725, opened 2025-03-26
man: clarify what the power class numbers mean for 802.3at, in dot3

systemcrash:


vincentbernat:
To avoid the space before comma:
```
.Ar class-0 ,
```

vincentbernat:
Thanks!

PR #735, opened 2025-05-23
Fix memory leaks in C++ wrapper

bitpwr:
`LldpCtl::GetInterfaces()` and `LldpAtom::GetAtomList()` do not decrement references for atoms retrieved from `lldpctl_atom_get()` and `lldpctl_get_interfaces`.

Added calls to `lldpctl_atom_dec_ref()`.

### DIFF
--- a/src/lib/lldpctl.hpp
+++ b/src/lib/lldpctl.hpp
@@ -235,6 +235,7 @@ class LldpAtom {
 		{
 			list.emplace_back(atom, true, conn_);
 		}
+		::lldpctl_atom_dec_ref(it);
 
 		return list;
 	}
@@ -364,7 +365,8 @@ class LldpCtl {
 
 	std::list<LldpAtom> GetInterfaces() const
 	{
-		const auto &it { ::lldpctl_get_interfaces(conn_.get()) };
+		lldpctl_atom_t *it;
+		CHECK_LLDP_P(it = ::lldpctl_get_interfaces(conn_.get()), conn_.get());
 
 		std::list<LldpAtom> list;
 		lldpctl_atom_t *atom;
@@ -372,6 +374,7 @@ class LldpCtl {
 		{
 			list.emplace_back(atom, true, conn_);
 		}
+		::lldpctl_atom_dec_ref(it);
 
 		return list;
 	}


src/lib/lldpctl.hpp
```diff
@@ -364,14 +365,16 @@ class LldpCtl {
 
 	std::list<LldpAtom> GetInterfaces() const
 	{
-		const auto &it { ::lldpctl_get_interfaces(conn_.get()) };
+		lldpctl_atom_t *it;
+		CHECK_LLDP_P(it = ::lldpctl_get_interfaces(conn_.get()), conn_.get());
```
rbu9fe, line 369: `CHECK_LLDP_P` will throw when a `nulltpr` is returned, but that seems to be a valid case. So better remove `CHECK_LLDP_P`.

src/lib/lldpctl.hpp
```diff
@@ -235,6 +235,7 @@ class LldpAtom {
 		{
 			list.emplace_back(atom, true, conn_);
 		}
+		::lldpctl_atom_dec_ref(it);
```
rbu9fe, line 238: I'll have to test these changes, if I remember correctly I once had an issue when decrementing the refcount. But I agree that normally we should decrement it as `lldpctl_atom_get` incremented it.

src/lib/lldpctl.hpp
bitpwr: Sure, can remove `CHECK_LLDP_P`, just thought we should be consistent. `lldpctl_atom_dec_ref` is safe to call with `nullptr` so I think that can stay?

src/lib/lldpctl.hpp
```diff
@@ -235,6 +235,7 @@ class LldpAtom {
 		{
 			list.emplace_back(atom, true, conn_);
```
rbu9fe, line 236: Here we need to add a reference to `this` to extend its lifetime as long as the atoms in the list are alive:
```suggestion
			list.emplace_back(atom, true, conn_, std::make_unique<LldpAtom>(*this));
```

src/lib/lldpctl.hpp
rbu9fe: I did some leack check tests and the changes are good.

src/lib/lldpctl.hpp
bitpwr: Removed the `CHECK_LLDP_P` macro.

src/lib/lldpctl.hpp
bitpwr: Yes, you are right, if the children cannot live without its parent atom. Updated as you suggested.

PR #740, opened 2025-07-28
add support for vlan-advertisements configuration

ugoldfeld:
New configurations to exclude a specific VLANs from being advertised via LLDP on a given interface.
The purpose is to filter out internal VLANs that are irrelevant for publishing. 
See discussion in https://github.com/lldpd/lldpd/discussions/737

Then added commands are:
configure [ports ethx [,..]] lldp vlan-advertisements pattern _pattern_.
unconfigure [ports ethx [,...]] lldp vlan-advertisements pattern.

The _pattern_ option is similar to the pattern in configure system interface.

For example:
publish all VLANs except 23,24:
configure ports eth0 lldp vlan-advertisements pattern "*,!23,!24"
publish only VLANs  23,24:
configure ports eth0 lldp vlan-advertisements pattern "23,24"

### DIFF
--- a/src/lldpd-structs.c
+++ b/src/lldpd-structs.c
@@ -211,6 +211,8 @@ lldpd_port_cleanup(struct lldpd_port *port, int all)
 			port->p_chassis->c_refcount--;
 			port->p_chassis = NULL;
 		}
+		free(port->p_vlan_advertise_pattern);
+		port->p_vlan_advertise_pattern = NULL;
 #ifdef ENABLE_CUSTOM
 		lldpd_custom_list_cleanup(port);
 #endif


src/client/conf-lldp.c
```diff
@@ -756,6 +804,22 @@ register_commands_configure_lldp(struct cmd_node *configure,
...
+			"Set vlan pattern", NULL, NULL, NULL),
+		NULL, "vlan pattern (comma-separated list of wildcards)",
+		NULL, cmd_store_env_value, "vlan-pattern"),
+		NEWLINE, "Set active vlan pattern", NULL, cmd_set_vlan_pattern, NULL);
+
+	struct cmd_node *unconfigure_vlan_advertisements =
+	    commands_new(unconfigure_lldp, "vlan-advertisements", "Unconfigure vlan address advertisements", NULL, NULL, NULL);	
+	commands_new(commands_new(unconfigure_vlan_advertisements, "pattern",
+		"Delete any vlan pattern", NULL, NULL, NULL),
+		NEWLINE, "Clear vlan pattern", NULL, cmd_unset_vlan_pattern, NULL);
```
vincentbernat, line 821: Please, change `vlan` to `VLAN` for descriptions.

src/client/conf-lldp.c
```diff
@@ -413,6 +413,54 @@ cmd_vlan_tx(struct lldpctl_conn_t *conn, struct writer *w, struct cmd_env *env,
...
+    const void *arg)
+{
+	log_debug("lldpctl", "set vlan pattern");
+	lldpctl_atom_t *port;
+	const char *name;
+
+	const char *value = cmdenv_get(env, "vlan-pattern");
+	
+	log_info("lldpctl", "vlan-pattern set to new value %s",
+	    value ? value : "(none)");
```
vincentbernat, line 427: `VLAN pattern set to ...`

And you can drop the log_debug at the beginning since not much happens between the two.

vincentbernat (changes requested):
Thanks! It looks OK. What is missing:
- manual page should be updated to document this new configuration setting
- also add an example of invocation in `tests/lldpcli.conf`
- a test would be nice (in `tests/integration/test_dot1.py`)

src/client/conf-lldp.c
ugoldfeld: done

ugoldfeld:
Great! Do you have an estimate to when there will be a new Tag?

vincentbernat:
1.0.20 was released today.

PR #750, opened 2025-10-26
lldpd: fix leak in levent_snmp_read by cleaning up netsnmp_large_fd_set

MegaManSec:


vincentbernat:
Not needed when using only `FD_SETSIZE`, but it can't hurt.

PR #770, opened 2026-03-05
client: add vendor-specific organizational TLV display

ciroiriarte:
**Note**: my apologies, this is a re-attempt for https://github.com/lldpd/lldpd/pull/769 (git-related L8 issues)

---

Add config-driven interpretation of organizational TLVs (Type 127) in lldpcli. TLVs from vendors whose OUI is not natively handled are normally shown as raw hex. This change introduces a three-tier lookup:

1. Code handlers (compiled) for complex TLVs like Cisco ACI SubType 207
2. INI-style config files for simple flat TLVs (field type + label)
3. Hex fallback (unchanged default behavior)

Config files are read from SYSCONFDIR/lldpd.d/org-tlvs.conf.d/*.conf and support field types: uint8, uint16, uint32, string, ipv4, mac, hex. Multi-field definitions with named labels are supported.

Ship Cisco ACI and Juniper TLV definitions by default. This is entirely client-side — no daemon, IPC, or struct changes.

New files:
  src/client/org_tlv_handler.{h,c} — code handler registry
  src/client/org_tlv_handler_cisco.c — Cisco ACI handler (SubType 207)
  src/client/org_tlv_conf.{h,c} — INI parser, config loading, display
  src/client/org-tlvs.conf.d/ — shipped config files

Modified files:
  src/client/display.c — three-tier lookup in display_custom_tlvs()
  src/client/lldpcli.c — init handlers and load config at startup
  src/client/Makefile.am — add new sources, SYSCONFDIR, install target
  src/client/lldpcli.8.in — document ORGANIZATIONAL TLV DISPLAY section
  NEWS — changelog entry

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,12 @@
+lldpd (1.0.23)
+ * Changes:
+   + Add vendor-specific organizational TLV display via config files and code handlers
+     - Config files in SYSCONFDIR/lldpd.d/org-tlvs.conf.d/*.conf define how to interpret TLV payloads
+     - Supported field types: uint8, uint16, uint32, string, ipv4, mac, hex
+     - Code handlers for complex TLVs (e.g. Cisco ACI SubType 207)
+     - Ships with Cisco and Juniper TLV definitions
+     - Client-side only, no daemon changes
+
 lldpd (1.0.22)
  * Fix:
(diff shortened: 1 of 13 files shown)


ciroiriarte:
adding requested tests

PR #779, opened 2026-03-17
liblldpctl: Fix const correctness of the receive callback

bartelsielski:
In order to write data into the buffer provided to the receive callback, it needs to be mutable.


vincentbernat:
Thanks! This looks correct. I don't remember why I put a `const` here.

PR #786, opened 2026-05-09
Various fixes from Claude Code

vincentbernat:
Except for commit 083ad7986da8 (unprivileged and I don't think this is easily exploitable since this is an out of bound read of 1 byte) and commit b821cf1371a0 (short race condition making the socket accessible to anyone, but without privileges), everything else is more defense in depth or fixing non-security related bugs (memory leak).


vincentbernat:
NetBSD 9.4 is broken for unrelated reasons.

PR #787, opened 2026-05-12
Fix heap OOB read in VLAN decapsulation memmove

TristanInSec:
## Summary

In `lldpd_decode()`, the VLAN decapsulation `memmove` shifts frame data 4 bytes left starting at offset `2*ETHER_ADDR_LEN`. The source pointer is correctly offset by +4, but the length argument uses the full remaining frame length (`s - 2*ETHER_ADDR_LEN`) instead of accounting for the 4-byte shift (`s - 2*ETHER_ADDR_LEN - 4`).

When the received frame fills the hardware MTU allocation exactly, the `memmove` source range extends 4 bytes past the end of the heap buffer.

## Fix

Subtract the 4-byte VLAN tag size from the `memmove` length, matching the source offset adjustment.

### DIFF
--- a/src/daemon/lldpd.c
+++ b/src/daemon/lldpd.c
@@ -574,7 +574,7 @@ lldpd_decode(struct lldpd *cfg, char *frame, int s, struct lldpd_hardware *hardw
 		/* VLAN decapsulation means to shift 4 bytes left the frame from
 		 * offset 2*ETHER_ADDR_LEN */
 		memmove(frame + 2 * ETHER_ADDR_LEN, frame + 2 * ETHER_ADDR_LEN + 4,
-		    s - 2 * ETHER_ADDR_LEN);
+		    s - 2 * ETHER_ADDR_LEN - 4);
 		s -= 4;
 	}
 


vincentbernat:
Thanks!

TristanInSec:
> Thanks!

Hi @vincentbernat, you're welcome!
Btw, could we assign a CVE via GHSA since the fix is merged? Thanks.
Best,
Tristan

vincentbernat:
Out of bound read on the heap seems also harmless. The process does not have any private information to exfiltrate. How does requesting a CVE via GHSA works?

TristanInSec:
Thanks for merging!

To answer your question about GHSA CVEs: GitHub can assign a CVE through a Security Advisory on the repository. The process is:

1. Enable "Private vulnerability reporting" in repo Settings > Code security and analysis
2. I create a draft Security Advisory describing the vulnerability (CWE, CVSS, affected versions, fix commit)
3. You review and publish it
4. GitHub (as a CVE Numbering Authority) automatically assigns a CVE and publishes it to the CVE List

It looks like private vulnerability reporting is currently disabled on the repo -- if you toggle it on, I can file the draft advisory right away.

vincentbernat:
OK, it's enabled.

PR #789, opened 2026-05-27
fix: Send termination message on interface removal

rbu9fe:
According to IEEE802.1AB a TX_SHUTDOWN_FRAME shall be sent when when the admin status of the interface is disabled. So far that was only done when the daemon is shutting down.

This change now sends a termination message as well when the interface is being removed from the interface pattern.


vincentbernat:
Thanks!

PR #792, opened 2026-06-12
Fix mixup between single and dual signature in strings.

Ekmansoft:
The strings describing POE 802.3bt single and dual signature PDs was mixed up when decoding from table 802.3-2022 table 79-14.

### DIFF
--- a/src/lib/atoms/dot3.c
+++ b/src/lib/atoms/dot3.c
@@ -77,8 +77,8 @@ static struct atom_map port_dot3_power_pse_status_map = {
 	.map = {
 		{ 0, "unknown" },
 		{ 1, "2-pair powering" },
-		{ 2, "4-pair powering dual-signature PD" },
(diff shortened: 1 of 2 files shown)


vincentbernat:
I think `port_dot3_power_pd_status_map` is another mistake:

image

Ekmansoft:
Hmm, if needed, I could include this as well, but I'm not sure whether this is correct in the spec, or if its a typo. 

To me it does not make sense to say we have a dual signature PD powered by two pairs... 

I would like to hold off on this change for now. To me it looks like a typo in the specs. 
I have sent a change request to IEEE.

Ekmansoft:
I asked a guy at Sifos who has been working a lot with the compliance testing of this. He thought the wording in the specs was probably correct, so maybe I should include that part too in the change.

Edit: It is included in in the current PR.

